PR #139, opened 2018-01-06
Codec generated tool. Additional setters is added

alx-ut:
1. The generation of getters/setters by Enum is added to Decoder/Encoder generation tool.
For ex.
Encoder:
```java
    public NewOrderSingleEncoder ordType(OrdType value)
    {
        return ordType(value.representation());
    }
```
Decoder:
```java
    public OrdType ordTypeAsEnum()
    {
        return OrdType.decode(ordType);
    }
```
2. Type of representation of char type in generated Enum is changed from 'int' to 'char'
3. DictionaryAcceptor: 
- added default empty implementation to method
- added onUnknownMessage(int messageType) with empty default implementation

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/CodecGenerationTool.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/CodecGenerationTool.java
@@ -59,17 +59,18 @@ public final class CodecGenerationTool
         final PackageOutputManager parent = new PackageOutputManager(outputPath, PARENT_PACKAGE);
         final PackageOutputManager decoder = new PackageOutputManager(outputPath, DECODER_PACKAGE);
 
-        final EnumGenerator enumGenerator = new EnumGenerator(dictionary, parent);
+        final EnumGenerator enumGenerator = new EnumGenerator(dictionary, PARENT_PACKAGE, parent);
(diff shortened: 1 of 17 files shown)


mjpt777:
Can you provide a description of the what need this PR addresses?

RichardWarburton:
I'm happy with most of these changes, But I would prefer to not have a default empty method for every acceptor method. If you want to have an easy way of overriding them I would prefer to generate an implementation class for the acceptor with empty implementatons of every method that someone could extend.

alx-ut:
Agree. 
default keyword is removed from method signature.
Class DefaultDictionaryAcceptor with empty implementations is generated.

alx-ut:
@mjpt777 :
The purpose of PR is to simplify usage of generated fix encoder/decoder:
- It's allow to use:
```java
newOrderSingleEncoder.ordType(OrdType.LIMIT) 
OrdType ordType = newOrderSingleDecoder.ordTypeAsEnum() 
```
invocations instead of 
```java
newOrderSingleEncoder.ordType((char)OrdType.LIMIT.representation()) 
OrdType ordType = OrdType.decode(newOrderSingleDecoder.ordType())
```
- DefaultDictionaryAcceptor  generation allow to override only desired message type acceptor instead of all one.

RichardWarburton:
Thanks for the contribution.

PR #144, opened 2018-05-29
Parsing repeating group elements using iteration instead of recursion

AndreyNudko:
Fixes StackOverflow when parsing repeating groups with 40K+ repeats (very real SecurityList in fixed income world)
Andrey & Stephen


RichardWarburton:
Thanks for the PR @AndreyNudko - merged.

PR #145, opened 2018-05-29
Improved dictionary parser to handle idiosyncrasies of certain vendor specs.

AndreyNudko:
This PR fixes two issues found when generating code from a real vendor FIX spec: -
  - using enum descriptions to generate Java identifiers produces uncompilable code if the description contains  illegal characters; the solution is to normalise the identifiers by translating illegal characters into underscore, and by prepending an underscore if needed;
  - resolve the situation of a duplicate field on a message, where the field was included once via a component and a second time directly; the solution is to use just a single copy of the field.

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/DictionaryParser.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/DictionaryParser.java
@@ -28,11 +28,16 @@ import javax.xml.xpath.XPath;
 import javax.xml.xpath.XPathExpression;
 import javax.xml.xpath.XPathExpressionException;
 import javax.xml.xpath.XPathFactory;
+import java.io.IOException;
 import java.io.InputStream;
+import java.io.UncheckedIOException;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.Objects;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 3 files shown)


RichardWarburton:
Thanks for this PR @AndreyNudko - looks good so I've merged it.

AndreyNudko:
Thanks @RichardWarburton - I'm actually starting to have doubts now whether this was a good change; perhaps specifying field on the message could be used to override one from component, e.g. to change required/optional flag?
Also, what might be more useful for us, is being able to use  CodecGenerationTool as a helper rather than command line utility - so another class can load and tweak dictionary as it needs. I may send another patch for that.

RichardWarburton:
Hmm, composing fields like that is a bit of a difficult challenge. I've considered what happens when you try to compose an identical required and optional field and I suspect that what most people want is for the field to become optional. Perhaps when the code is sanitizing the fields it should log out when a duplicate is being removed? That offers more transparency for a user and would allow them to customise things.

When you say customise it - do you mean be able to modify the resulting Dictionary? If so, I'd be interested to understand what advantages that has over just modifying the input XML file - if you're invoking the commandline tool then that is under your control.

AndreyNudko:
Right, our original problem was that we've got dictionary from vendor (which I suspect don't eat their own dog food, i.e. don't use it to generate code); so we tried the approach of adapting loader to handle that, and that lead to idea of decoupling loader and generator. But maybe it's better just to keep our sanitized copy of XML and occasionally merge changes from vendor file - the maintenance cost would probably be about the same. At the moment I can't say for sure, as we're just starting experimenting with code generation. As we'll use it for more FIX flavours, it should become clear if there is a good reason to split CodecGenerationTool.
In the meantime, I'll leave it up to you to decide if field deduplication logic is a good thing to have  - maybe it's a bit too 'smart'.

PR #155, opened 2018-06-04
If repeating group is not present just return an iterator whose .hasNext() is false

tom-smalls:
In reference to https://github.com/real-logic/artio/issues/152

Also introduces repeating groups implementing Iterable so "foreach" can be used


RichardWarburton:
Hi @tom-smalls - thanks for the PR - looks good.

PR #158, opened 2018-06-07
Add code style and inspection warnings for intellij

tom-smalls:
It makes it easier to resolve checkstyle failures


RichardWarburton:
Thanks for sending this, but I'd prefer it if we didn't merge in the intellij config. It means we have multiple style configuration things to maintain moving forward. Intellij does have a checkstyle plugin (https://plugins.jetbrains.com/plugin/1065-checkstyle-idea) that can be configured to use the checkstyle file from this project and will give you in IDE warnings when the code breaches these rules.

tom-smalls:
No worries.

Would it be possible for you to make the project's intellij config available?

Issue #153, opened 2018-06-03
Add an AsciiSequenceView


RichardWarburton:
This should be a view over an underlying buffer that can be used to flyweight a buffer or take a copy of a range. They could be passed in, rather than allocated and returned, and modified to have the appropriate range in the underlying buffer. These could work with SBE to avoid he allocation and do a straight copy in the most efficient way. To be used by the generated codecs instead of storing a char[] and length everywhere.

tom-smalls:
Would it make more sense for the `AsciiSequenceView ` to live in agrona rather than artio? This would allow for better/easier integration with SBE, plus we can think of usages for this outside FIX parsing.

tom-smalls:
We've created a pull request for introducing a AsciiSequenceView. We were wondering if this should be togglable in the generator, where either AsciiSequenceView methods or the array backed string fields are generated.

We haven't added the AsciiSequenceView for encoding. We thought this might be something that is useful for providing a view on buffer backed strings to allow for a more efficient copy, but if we did this we weren't sure of the best approach; holding onto a reference of a AsciiSequenceView until the Fix message is encoded doesn't feel right.

RichardWarburton:
Only thing missing for this issue is the AsciiSequenceView for the Encoder.

tom-smalls:
I just remembered why we didn't implement this for encoders (well, read https://github.com/real-logic/artio/issues/154 again)

``
We haven't added the AsciiSequenceView for encoding. We thought this might be something that is useful for providing a view on buffer backed strings to allow for a more efficient copy, but if we did this we weren't sure of the best approach; holding onto a reference of a AsciiSequenceView until the Fix message is encoded doesn't feel right.
``
Do you have a view regarding implementation?

RichardWarburton:
I'm not sure there's an alternative really other than that approach if we're going to avoid copying. Do you have any ideas or suggestions @mjpt777  ?

mjpt777:
I'm happy for the `AsciiSequenceView` to move to Agrona if that helps.

mjpt777:
I've added an `AsciiSequenceView` to Agrona with this commit. https://github.com/real-logic/agrona/commit/fb24b881f071e4e9a011c1fd2819aa554bc72dad

RichardWarburton:
I'm closing this issue as it's been resolved.

PR #160, opened 2018-06-07
[Tom/Zam/Andrey] #153 Add AsciiSequenceView

tom-smalls:
This can be passed into a decoder to provide:
* a view of the underlying buffer for a string-based field
* avoid allocation
* allows for more efficient copying of data


RichardWarburton:
Thanks for the PR. I'm a bit confused from just reading the diff. Is it required to use this for the encoding and decoding of String in your PR? I'd prefer it to not be required to be honest.

tom-smalls:
This change just adds a new method for decoding strings that allows the user to pass in `AsciiSequenceView`.

We were wondering if this should be togglable in the generator, where either AsciiSequenceView methods or the array backed string fields are generated.

We haven't added the AsciiSequenceView for encoding. We thought this might be something that is useful for providing a view on buffer backed strings to allow for a more efficient copy, but if we did this we weren't sure of the best approach; holding onto a reference of a AsciiSequenceView until the Fix message is encoded doesn't feel right.

PR #161, opened 2018-06-07
Bug fix: enum validation

tom-smalls:
We discovered an issue with the validation of `MultipleCharValue` and `MultipleStringValue` / `MultipleValueString`, where if more than one enum value was included in the FIX value (they are delimited by spaces) then the validation would fail as it would just take the first char in the former case or treat the FIX value as a single string in the latter case.

As part of the change to fix this we moved enum validation into the generated enum classes rather than it being done in the message decoder to handle the data types enums are backed by more easily. This required updating the underlying `CharArrrayMap` to take an offset as well as a char array and length.

We've also tried to make use of the generated `Constants` class for associating FIX tag numbers with their meaning. This meant having to move the `Constants` class up a level and to move the normalisation of FIX.4.4 repeating groups up a level from `Group` to `DictionaryParser`.


tom-smalls:
We believe this is also an issue for the `decode` method on `MultipleCharValue` and `MultipleStringValue` / `MultipleValueString` enum classes. However, the generated message decoders do not create `asEnum` methods for these values. A user would have to call the enum `decode` method themselves so we have left it for the moment.

We were thinking of having the `decode` method return an `Iterator` of enums as a potential fix.

RichardWarburton:
Hi @tom-smalls, thanks for sending the PR. In general it looks good. Can you confirm that you've run the tests in https://github.com/real-logic/fix-integration on this code?

Also - if it's possible in future I'd really appreciate trying to keep these PRs single purpose and as small as possible. It makes it easier (and quicker) for me to review and merge them.

tom-smalls:
@RichardWarburton I hadn't but I have now run the fix-integration tests.

All the tests passed but needed to change the import for `Constants` in `OtfParsesBytesFromEncoderTest` and `OtfParserQuickFixIntegrationTest` from `import static uk.co.real_logic.artio.decoder.Constants.*;` to `import static uk.co.real_logic.artio.Constants.*;`. Would you like me to raise a PR for this?

Yes, apologies for the multiple changes in the PR. We came across the validation bug when looking at reusing `Constants` some more

RichardWarburton:
Yes - raising that PR would be most appreciated.

RichardWarburton:
I think I need you to push something to resolve the conflict here Tom.

RichardWarburton:
Great - thanks!

Issue #157, opened 2018-06-06
Provide a sentinel for missing enum values rather than throwing an IllegalArgumentException


RichardWarburton:
Also improve consistency in enum behaviour:

Incidentally, we have found possibly unintentional inconsistent behaviour between String represented enums and the char and int represented ones. The decode() method that takes a char[] returns null when an unknown enum value is found, and the others throw an IllegalArgument. See the generated return statement in EnumGenerator::optionalCharArrayDecode

PR #162, opened 2018-06-08
#157 Change sentinel values for missing INT and LONG to min values

merliecat:
The argument is that sentinels of -1 are more likely to be used as actual values

This is in preparation for returning sentinel values for enumerations


RichardWarburton:
Thanks @merliecat This PR's title says it only changes missing int and long sentinels, but actually the PR has a load of code in for resetting strings. Is that correct?

merliecat:
Nope, sorry, an additional commit was not meant to be pushed. I'll close this and create another PR.

PR #163, opened 2018-06-09
#157 Change sentinel values for missing INT and LONG to min values

juddgaddie:
The argument is that sentinels of -1 are more likely to be used as actual values

This is in preparation for returning sentinel values for enumerations


RichardWarburton:
Thanks, but again like the previous PR this is described as a change to missing values but is actually a lot of other things related to Strings.

juddgaddie:
ok, I now see what you are referring to.

The production code is only changing sentinels for int and long. However included in this patch is a set of tests asserting existing behaviour for required fields for each type when their values are reset()

i.e. for string, integer, char, decimal, and all enum variants named in form:
shouldNotRetain**TYPE**FromPreviousMessagesForRequiredFieldsWhenReset

RichardWarburton:
Ahh, thanks for clarifying.

PR #164, opened 2018-06-09
[Stephen/Judd] #157 return sentinel values for NULL_VAL or UNKNOWN_RE…

### LINKED ISSUE #157: Provide a sentinel for missing enum values rather than throwing an IllegalArgumentException

juddgaddie:
…PRESENTATION for enums instead.

Required fields are assumed to be present and therefore will only return UNKOWN_REPRESENTATION when called. Optional fields will return either NULL or UNKOWN_REPRESENTATION.

This closes #157

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
@@ -32,6 +32,7 @@ import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 import static uk.co.real_logic.artio.dictionary.generation.AggregateType.*;
 import static uk.co.real_logic.artio.dictionary.generation.ConstantGenerator.sizeHashSet;
+import static uk.co.real_logic.artio.dictionary.generation.EnumGenerator.NULL_VALUE_NAME;
 import static uk.co.real_logic.artio.dictionary.generation.Exceptions.rethrown;
 import static uk.co.real_logic.artio.dictionary.generation.GenerationUtil.constantName;
 import static uk.co.real_logic.artio.dictionary.generation.GenerationUtil.fileHeader;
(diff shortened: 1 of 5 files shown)


RichardWarburton:
Hi Judd thanks for the PR. I think we should try and keep things as simple and clear as possible here. Just thinking things through here so correct me if I'm wrong on anything.

1) We've been using the term MISSING_VALUE for the value of a field that is parsed and isn't present. It looks like NULL_VALUE means missing here. If that's correct then I think it should be called MISSING_VALUE for consistency, and I think there should be constants on the CodecUtil class for the missing value constants so that they can be referenced by user code.

2) Do we really need the distinction between MISSING_VALUE and UNKNOWN_REPRESENTATION?

merliecat:
Richard,
Having considered things we agree with your point 2, that it is overkill to have two sentinel values.  So, we're happy to reduce down to just one.
However, we would like to use the name UNKNOWN_REPRESENTATION for the sentinel.
Our impetus for introducing the sentinel was to deal with the case a value has been provided, but it does not correspond to any of the valid enum representations.
The current behaviour is to throw this exception: -
`throw new IllegalArgumentException("Unknown: " + representation);`
so we based the name on be semantics of the situation to be represented.
What do you think?

RichardWarburton:
Oh I see. Yes - a sentinel called UNKNOWN_REPRESENTATION would be fine for the case that there's a value there, but its not a valid enum. I think we should be sticking with MISSING_X for cases where the value is genuinely missing, eg: MISSING_INT where it isn't an enum. Is there any overlap here? For example what would I get if I've got a missing field that's an enum?

mjpt777:
`UNKNOWN` would be more consistent with SBE. In SBE we use `UNKNOWN` for a value that is unknown and maybe added later, and `NULL_VAL` to represent an optional value that is not present.

RichardWarburton:
I'm happy to go with SBE consistency.

merliecat:
We've updated this PR now based on discussions.

Issue #150, opened 2018-05-30
Log when field deduplication happens in codec tool


RichardWarburton:
As per discussion in #145 .

PR #165, opened 2018-06-12
[Tom] #150 update the generator to fail on duplicate fields

tom-smalls:
The FIX specification states:

"A tag number (field) should only appear in a message once. If it
appears more than once in the message it should be considered an error
with the specification document. The error should be pointed out to
the FIX Global Technical Committee."

Failing the code generation informs the user the FIX spec they are
using is "invalid"; attempting to use the generated codecs may
result in unexpected behaviour

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/DictionaryParser.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/DictionaryParser.java
@@ -15,11 +15,16 @@
  */
 package uk.co.real_logic.artio.dictionary;
 
-import org.agrona.Verify;
-import org.w3c.dom.*;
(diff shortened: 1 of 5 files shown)


tom-smalls:
@AndreyNudko previously raised a PR where duplicate fields are removed, but this did not check whole "tree" of a message.

We have experienced unexpected behaviour where fields have been declared multiple times for a message, and this combined with the FIX specification stating fields should not be duplicated, has led us to wanting to have the code generation fail in these conditions.

Currently the https://github.com/real-logic/fix-integration will fail as QuickFixJ have duplicate fields declared in their fix44.xml specification. We have raised a PR to resolve this https://github.com/quickfix-j/quickfixj/pull/197

tom-smalls:
The QuickFix PR has been merged and https://github.com/real-logic/fix-integration is now passing with this change

PR #166, opened 2018-06-12
[Tom] improve toString() for message with repeating group

tom-smalls:
Currently, the toString() for a decoder checks to see if a repeating
group is null or not to determine what to output. This can be misleading
in the following scenario:

1) Parse a message that does not contain the repeating group
2) Parse a message that has a repeating group with multiple elements
3) Parse a message that does not contain the repeating group

In the case of the messages 1) and 3) being the same, they will produce
different output on toString() as the repeating group is no longer null
due to the parsing of message 2).


RichardWarburton:
Looks good, thanks.

PR #168, opened 2018-06-13
Clean up generation of GROUP_FIELDS for decoders.

tom-smalls:
Given a the following message definition:
```
<message name="NewOrderSingle" msgtype="D" msgcat="app">
  <field name="ClOrdID" required="Y"/>
  <component name="Members" required="N"/>
</message>
<component name="Members">
<group name="NoMemberIDs" required="N">
  <field name="MemberID" required="Y"/>
  <field name="Text" required="N"/>
</group>
</component>
```
The NewOrderSingleDecoder has ClOrdID, MemberID, and Text inside it's GROUP_FIELDS IntHashSet.
The MembersGroupDecoder has  MemberID, and Text inside it's GROUP_FIELDS IntHashSet.

NewOrderSingleDecoder only needs to know about ClOrdID as MembersGroupDecoder deals
with the other fields (and whether they are known).

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
@@ -252,7 +252,11 @@ public class DecoderGenerator extends Generator
 
         if (aggregate.containsGroup())
         {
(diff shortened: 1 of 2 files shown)


tom-smalls:
For context, I am working towards having the `validate` method on repeating groups being called when `validate` is called on a Message.

RichardWarburton:
Thanks.

PR #169, opened 2018-06-13
Improve validate method generated by message decoders.

tom-smalls:
Currently, the `validate()` on a message decode only checks the child
fields of the message and none of it's descendants. This means the user
of a message decoder would have to call `validate()` on each group 
entry whilst iterating over the group.

This change updates  `validate()` on a message decoders. It now iterates 
through any repeating group entries (if there are any)
and call `validate()` on each entry.


RichardWarburton:
Thanks.

PR #171, opened 2018-06-14
Change order field validation occurs.

tom-smalls:
At the moment the decoders check if a required field has been seen
before checking to see if any unknown fields have been visited. This can
be misleading in the case of an unknown field being seen before a
required field as the validation failure will refer to a missing
required field where the real failure is due to an unknown field being
present.

An unknown field can appear at the same time as a required field being
missing, but a required field being missing can't cause an unknown field 
failure to occur.


RichardWarburton:
Hi Tom, have you checked that this passes the fix integration test project? I'm not that fussed on the order of error reporting myself, but I'm sure I remember those quickfix integration tests making an assumption about the order of errors.

tom-smalls:
Yep.

What I wasn't sure about is how tolerant decoders should be of unknown fields. If a decoder comes across a field it doesn't know about it, should it stop parsing the message or step over that field and continue (providing some sort of feedback to the user of the decoder)? I could imagine a venue would want to reject but a client/bank would not necessarily want to do this as the source of truth of what the message encoding *should* be is the venue.

RichardWarburton:
I think if validation is switched on it should fail, if validation is switched off it should be fine.

We've discussed having validation that can be switched on or off on a per validation-check property, but its never been a priority for anyone.

tom-smalls:
Happy to accept this and could raise a separate issue about changing how validation is turned on or off (if that's something worth pursuing)?

RichardWarburton:
@tom-smalls - am I right in thinking that "I think if validation is switched on it should fail, if validation is switched off it should be fine." is currently true?

(I hadn't accepted the PR because I was under the impression from your question that there was still work to be done on that front.)

RichardWarburton:
Ok, thanks @tom-smalls

RichardWarburton:
Yeah. So unfortunately when I did a release the other week I forgot to bump the version of Artio that fix-integration was depending upon - my bad. As a result when you ran this change of validation order code against fix-integration it ran the last 0.15-SNAPSHOT you had installed. The failing test is `14b_RequiredFieldMissing.def`.

Do you think this validation order is particularly important? If it is then I guess we can maintain our own version of `14b_RequiredFieldMissing.def`, otherwise I'd be inclined to revert this change. If you think the order of validation is important then would you be happy to submit a PR to the fix-integration project to make a local copy of `14b_RequiredFieldMissing.def` without the ambiguity that we hit and get it passing again?

tom-smalls:
I am not too familiar with the fix-integration code. Is ` 14b_RequiredFieldMissing.def` something that is defined by QuickFix?

I can see a few references to it in the code and happy to raise a PR if you are referring to updating FIX message definitions in `ApplicationMessageValidationTest` and `SessionMessageValidationTest`

RichardWarburton:
Yeah - the fix-integration project basically takes a bunch of integration tests that quickfix defines and runs them between an Artio instance and a quickfix instance. This gives us test coverage that Artio can actually talk to another FIX engine that's in use. `14b_RequiredFieldMissing.def` is one of their test scenarios. We have modified versions in `test-fix-44/src/test/resources/custom_definitions/fix44/` where the behaviour is a little different - you can add a modification to that in the Artio fix-integration project. Or maybe quickfix upstream would take a revised 14b that doesn't have the ambiguity.

tom-smalls:
I'll have a look at updating what QuickFix as it feels like wrong to have the ambiguity.

In the mean time, would it be best to revert the change due to the broken fix-integration tests?

RichardWarburton:
Right - I've pushed a revert on this.

tom-smalls:
Ok.

I will look into the QuickFix tests and try to update it before raising this PR again

tom-smalls:
After some investigation I found that the FIX message that has been altered in https://github.com/real-logic/fix-integration/pull/4 is the culprit for the failing test.

I ran fix-integration with and without the revert with this change and it passes both times; no changes to QuickFix code required.

Would you like me to raise another PR (assuming you're still happy to take this change)?

RichardWarburton:
Hi Tom, I'd be happy to take another PR and yes that sounds like the correct test.

PR #172, opened 2018-06-19
NPE when an unknown field reached in repeating group

tom-smalls:
In the following scenario:
 
* User receives a message for the first time, which contains a repeating group
* When parsing the repeating group entries, an unknown field is reached (not in the last group)
* Decoder for the repeating group exits to it's parent as it does not know how to handle the field without initialising it's `next` field
* As the parent is still inside the repeating group for-loop, it calls `decode` on the `next` element, resulting in an NPE

I believe this may have been introduced as part of https://github.com/real-logic/artio/pull/144 as the ownership of when a repeating group has finished decoding was split.

I was wondering if one of the following may be a better approach:
* lazily instantiating the `next` field by the parent
* just stop trying to decode the repeating group as new instances of the same repeating group decoder are not going to be able to parse this field either

However, when I tried these a lot of tests started to fail unexpectedly but this PR is fixing the symptom rather than the problem.

Thoughts?


RichardWarburton:
I'll accept this PR to fix the NPE.

In terms of your bigger question I think if you're trying to decode a group and it fails validation with validation switched on then just stopping is best. I think the general principle around validation is that having validation switched on should fail-fast where possible, but validation switched off should be heroic - ie try to ignore errors and keep going where possible.

PR #173, opened 2018-06-19
Reference Constants.ALL_FIELDS rather than copy.

tom-smalls:
As part of integrating with venues, we sometimes receive SecurityList messages with over 40,000 instruments within it's repeating group. This means we end up with over 40,000 copies of the allFields map that is roughly 8000 bytes each.

This commit changes the copy to a reference

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
@@ -105,8 +105,6 @@ public class DecoderGenerator extends Generator
 
     private String generateAllFieldsDictionary()
     {
-        final int hashMapSize = sizeHashSet(dictionary.fields().values());
-
         return intHashSetCopy(ALL_FIELDS, "Constants.ALL_FIELDS");
     }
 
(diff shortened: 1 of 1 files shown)


RichardWarburton:
Hmm, I can see the motivation for reduced memory consumption, but I think this PR would introduce a race condition.

If you have multiple threads decoding FIX messages at the same time within the same process then they share the static ALL_FIELDS value. The Agrona IntHashSet reuses its internal iterator instance so you've got multiple threads racing around the internal state on the iterator. I think if you run the ConcurrencyStressTest you'll see this problem. That's why we copy them at the moment.

Since the codec generator was written the IntHashSet in Agrona has been improved - so maybe there's a better way of re-using the iterator without having a race condition now. If not - I would suggest raising an issue on Agrona to get feedback on what the right approach is.

tom-smalls:
Can you point me at where the iterator is used for `allFields`? I'm not sure where it's used within the `decode`. I don't see a reason for the field to be `public` so could be made `private` if it's a concern users will access it

RichardWarburton:
Oh - you're right. It's only `missingRequiredFields` `unknownFields` that are iterated over. My apologies.

In that case - is it possible to remove the `allFields` field and just get all the codecs to refer to the `Constants.ALL_FIELDS` value? Looks like that would simplify even further.

tom-smalls:
Agreed, that's a much better approach.

I'll look at doing that tomorrow morning.

tom-smalls:
I have made this change now

PR #174, opened 2018-06-19
Remove dead unused fields

tom-smalls:
"unknownFields" and "alreadyVisitedFields" are used by a message for validating if a field is unknown or has been seen twice. For groups, these fields are populated but never used as both are used (in part) to determine the end of a repeating group entry. These fields are never queried at a group level so should just remove them


RichardWarburton:
Looks good, thanks.

PR #180, opened 2018-07-05
[Tom/Andrey] call reset on all repeating group entries

tom-smalls:
Currently, reset is only being called on the first entry of a repeating group, which can result in corrupt/invalid data when future messages are parsed. If you receive a message with 3 entries in a repeating group with all fields populated, and then parse another message (after calling reset) with 3 entries in the repeating group but those entries do not have optional fields set, then you will send up with a corrupt message as the optional fields are still set from the first message that was parsed.

This commit iterates over entries in a repeating group and resets all of them.

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
@@ -177,6 +177,40 @@ public class DecoderGenerator extends Generator
         return Decoder.class;
     }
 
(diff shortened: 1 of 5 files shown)


tom-smalls:
@RichardWarburton any thoughts on this pull request? I alternative approach could be to "lazily" reset repeating group entries when required to reduce the "cost" of reset but I haven't experimented with how successful such an approach would be

RichardWarburton:
Hi @tom-smalls,

My apologies for not replying to this earlier, totally slipped my mind. The PR looks good, thanks.

PR #189, opened 2018-10-30
Ignore unknown fields for message when parsing

tom-smalls:
This PR changes the default behaviour of the decoders to pass over an unknown field for a given message; original behaviour can be retained by setting the property `fix.codecs.reject_unknown_field` to `true`.

Previously, decoders would stop trying to parse a "component" or whole message if an unknown field was reached and validation was switched on, or an unknown field was reached and the field is known in the FIX spec (but not necessarily for the given message).

We have found that venues will add new FIX fields to messages without warning so being robust to this scenario is quite important.

The generated `HeaderDecoder` will not attempt to continue parsing a message if an unknown field is reached as it has the requirement of being used independently of any message so cannot rely on the context of "known" messages


RichardWarburton:
Thanks @tom-smalls . now I've merged your PR on fix-integration, does it pass with these changes committed?

RichardWarburton:
Otherwise it looks good.

tom-smalls:
Yes it does. `fix-integration` will run and reject when an unknown field is received. I am not sure if QuickFIX have test cases for skipping over unknown fields, they might be worth adding to the test suite if so.

RichardWarburton:
It looks like you've added that test coverage to the DecoderGeneratorTest so I think we're ok. I'd rather have the specifics of codecs tested at that level as they run faster and are easier to modify/debug.

PR #190, opened 2018-11-01
[Tom] NPE can occur when calling reset on decoder 

tom-smalls:
Whilst decoding a repeating group, if an unknown field (or field should appear outside of the repeating group) is reached before getting to the last entry, a NPE will occur if reset is called on the decoder as it will try and call reset on all repeating group entries, but because entries are lazily instantiated the last entries were missed as decoding finished prematurely.

This commit places a null guard in the reset iterator call for groups


RichardWarburton:
Hi @tom-smalls, thanks for the PR. Just having a look over it won't calling .next() inside the foreach loop advance the position twice for each iteration?

tom-smalls:
Yes, it's a bit strange. The `next` method being called actually just returns the value in the current elements "next" field, it doesn't actually move the iterator on

RichardWarburton:
I see. I guess ideally the hasNext() should the check and the next() should progress the iterator.

tom-smalls:
True, I'll have a look at it

On 1 Nov 2018 22:41, "Richard Warburton" <notifications@github.com> wrote:

I see. I guess ideally the hasNext() should the check and the next() should
progress the iterator.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/real-logic/artio/pull/190#issuecomment-435211500>, or mute
the thread

tom-smalls:
Looking at the generated code of a message, a generated reset method for a repeating group looks like:

```
public void resetMDEntriesGroup()
    {
        for (final MDEntriesGroupDecoder iterator : mDEntriesGroupIterator.iterator())
        {
            iterator.reset();
            if (iterator.next() == null)
            {
                break;
            }
        }
        noMDEntriesGroupCounter = 0;
        hasNoMDEntriesGroupCounter = false;
    }
```

The `iterator()` call is returning the decoder, not an iterator, so the `next()` method does not relate to iteration at all. Slightly confusing

RichardWarburton:
Ok, in that case the fix looks good. But we should probably rename that loop variable in the generated code to avoid future confusion.

tom-smalls:
Raised a PR for the rename

PR #192, opened 2018-11-05
Rename field inside repeating group reset method

tom-smalls:
rename field inside repeating group reset method from "iterator… to the name of the decoder class as iterator is quite a misleading name considering you are not interacting with an iterator


RichardWarburton:
Great, thanks.

PR #193, opened 2018-11-05
Handle undefined repeating group in FIX message

tom-smalls:
With the introduction of being able to skip fields, we need to be able to handle the case of an unknown repeating group in a message.

An unknown repeating group in a message will fail validation due to seeing the same field more than once as the message does not know how to deal with the field.

This fix will remove a tag from the alreadySeenFields Set for a given component, for whilst it has been seen, it has not been handled by this component


RichardWarburton:
Thanks @tom-smalls

PR #200, opened 2018-11-20
normalisation and toDouble() and fromDouble()

merliecat:
This pull request includes both the normalisation of DecimalFloats and the convertions to and from doubles.

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/fields/DecimalFloat.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/fields/DecimalFloat.java
@@ -43,8 +43,13 @@ import java.util.Arrays;
  */
 public final class DecimalFloat implements Comparable<DecimalFloat>
 {
-    public static final DecimalFloat MIN_VALUE = new DecimalFloat(Long.MIN_VALUE, 0);
-    public static final DecimalFloat MAX_VALUE = new DecimalFloat(Long.MAX_VALUE, 0);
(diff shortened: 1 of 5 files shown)


artio-codecs/src/main/java/uk/co/real_logic/artio/fields/DecimalFloat.java
```diff
@@ -189,4 +230,151 @@ public int compareTo(final DecimalFloat other)
...
+    private void setAndNormalise(final long value, final int scale)
+    {
+        this.value = value;
+        this.scale = scale;
+        normalise();
+    }
+
+    private void normalise()
+    {
+        final long valueCopy = value;
```
RichardWarburton: You only use those copies in the ArithmeticException whose method references them. Shouldn't the arithemtic exception throw the new values and you reset the fields to the copy that you started with?

Perhaps an easier way to write this code would be to perform all the operations on local variables and then only set them to the fields at the end if they are inside the limits.

RichardWarburton:
One comment aside, this looks good. Thanks a lot.

artio-codecs/src/main/java/uk/co/real_logic/artio/fields/DecimalFloat.java
merliecat: That's a fair comment.  I'll change to using local variables.

merliecat:
Sorry it took a few days to get back to this.
I've commit the change you requested to use local variables in the normalise() method.
Regards.

RichardWarburton:
Cheers @merliecat

PR #215, opened 2019-01-07
Additional methods onComponent interface

tom-smalls:
When an Aggregate is a FIX Component, an interface is generated to allow for code sharing when decoding FIX messages; however, not all fields are on this interface. This PR adds the ability to consume enum FIX values as an enum, String using AsciiSequenceView, and exposing the number field for a repeating group on the interface

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
@@ -538,7 +538,6 @@ public class DecoderGenerator extends Generator
     {
         groupClass(group, out);
         final Entry numberField = group.numberField();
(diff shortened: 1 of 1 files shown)


tom-smalls:
@RichardWarburton I am not sure what these failures are regarding? All the tests for the branch pass locally for me

RichardWarburton:
I see - Thanks for reporting the fix-integration situation - I'll take a look at that. The CI setup that I've added doesn't run fix-integration so far so it won't be that.

You should be able to see the failing tests at:

https://ci.appveyor.com/project/RichardWarburton/artio/builds/21438513/job/bcfvfab3g1sq2emi/tests

tom-smalls:
I messed up my branching! What I had locally is different to what is pushed to github

RichardWarburton:
Thanks @tom-smalls

PR #226, opened 2019-01-18
[Tom/Zam/Michael] Remove OP_CONNECT after connection succeeds.

tom-smalls:
If you do not remove OP_CONNECT the InitiatedChannelHandler can be called back in the case of a ConnectionResetByPeer. See https://github.com/netty/netty/issues/924 for a better explanation

We attempted to write a test for this but not sure how to write a test that creates a TCP connection in Artio. If you can point me in the right direction I am more than happy to do this


RichardWarburton:
Thanks for spotting this @tom-smalls, looks good.

PR #229, opened 2019-01-21
Provide a default DecimalFloat value in encoders.

tom-smalls:
Following the same pattern as byte array backed fields, instantiate all fields that are DecimalFloat with an instance (rather than null). This allows users to pass in the scale and value rather than a DecimalFloat when encoding


tom-smalls:
This isn't quite ready yet, please discard!

tom-smalls:
Ended up adding a new setter method for `DecimalFloat` fields on the encoders to ensure the "has" field for optional FIX fields is set appropriately

RichardWarburton:
Thanks @tom-smalls  - looks good.

Issue #227, opened 2019-01-21
Enhancement of class MutableAsciiBuffer method putAsciiFloat


ariksher-nex:
it seems that if decimalFloat represents value between 0 (exclusive) and 1 (exclusive)
the method putAsciiFloat will output the value **without** leading zero.
can be seen in test class **DecimalFloatEncodingTest** relevant entries : 
            {".995", 995L, 3}
            {"-.995", -995L, 3},
            {".6", 6L, 1},
            {".06", 6L, 2},
            {"-.6", -6L, 1},
            {"-.06", -6L, 2},
            {".92117125", 92117125L, 8},
            {"-.92117125", -92117125L, 8},
            {".92125", 92125L, 5},
            {"-.92125", -92125L, 5},
            {".00007875", 7875, 8},
            {"-.00007875", -7875, 8},

i think that need had behavior flag for the MutableAsciiBuffer to control whether to add this leading zero or not, the default would be without, just to preserve the existing behavior.
proposed flag name : **outputLeadingZeroForDecimalFloat**  (default value is false)

RichardWarburton:
Can you explain what the need is providing the leading zeros? As far as I know the FIX spec is fine with you omitting the leading zeros. I've not so far encountered any systems that don't accept this, however, if you have then I'd be interested to know of it.

I don't like the proposed solution to add another configuration option to control this behavior. If leading zeros is needed then it's probably simpler to just produce leading zeros everywhere, you're definitely against the FIX specification if you don't parse "0.1". I've been considering going down this route myself btw just to remove the risk of hitting a system that didn't support leading zeros.

ariksher-nex:
we're rewriting our system, and the old system does send leading zero, for example: "0.1"
so our customers expect it, currently not sure if they handle values without leading zeros (in this example ".1").

I suggested the solution of flag just to preserve your existing behavior,
if you don't like it, i'm with you, lets make it consistent and always have leading zero.

ariksher-nex:
yep, I will do so.

ariksher-nex:
BTW I encountered some issue, can you please take a look at issue #228

PR #230, opened 2019-01-21
Enhancement of class MutableAsciiBuffer method putAsciiFloat - leading zeros

ariksher-nex:
solves issue #227


RichardWarburton:
Thanks for the PR, but it has caused the build to fail - please run `./gradlew` and fix the checkstyle errors.

artio-codecs/src/main/java/uk/co/real_logic/artio/util/MutableAsciiBuffer.java
```diff
@@ -45,6 +47,30 @@
...
+        10_000_000_000L,
+        100_000_000_000L,
+        1_000_000_000_000L, // scale=12
+        10_000_000_000_000L,
+        100_000_000_000_000L,
+        1_000_000_000_000_000L, // scale=15
+        10_000_000_000_000_000L,
+        100_000_000_000_000_000L,
+        1_000_000_000_000_000_000L, // scale=18
+        //10_000_000_000_000_000_000L, // scale=19
```
RichardWarburton: Why is this commented out?

artio-codecs/src/test/java/uk/co/real_logic/artio/fields/DecimalFloatToStringTest.java
```diff
@@ -48,6 +48,8 @@ public void canEncodeDecimalFloat()
     {
         final DecimalFloat price = new DecimalFloat(value, scale);
 
-        assertEquals(input, price.toString());
+        // since toString of DecimalFloat does not return leading zeros, it fails
+        //assertEquals(input, price.toString());
```
RichardWarburton: Why does this PR just comment out asserts? The the output of the `toString()` method should be consistent with printing the number on the `AsciiBuffer`.

artio-codecs/src/test/java/uk/co/real_logic/artio/util/DecimalFloatEncodingTest.java
```diff
@@ -36,27 +36,92 @@
...
             {"1.00109125", 100109125, 8},
             {"-1.00109125", -100109125, 8},
+
+            // negative scale
+            {"2000", 2L, -3},
+            {"-2000", -2L, -3},
+
+            // edge values
+            {"0", 0, 0},
+            //{"0.00", 0, 2},
```
RichardWarburton: Why is this value commented out? Either it should pass or if it isn't a valid value then it should be removed.

artio-codecs/src/main/java/uk/co/real_logic/artio/util/MutableAsciiBuffer.java
ariksher-nex: cannot really represent the value of 10 * 10^18.
look at method **checkForLeadingZeroAndPutIfNeeded** and see that if scale is GE 19 there is special treatment.

artio-codecs/src/main/java/uk/co/real_logic/artio/util/MutableAsciiBuffer.java
```diff
@@ -617,6 +644,16 @@ private long calculateRemainderAndPutMinus(final int offset, final long value)
         }
     }
 
+    private int checkForLeadingZeroAndPutIfNeeded(final int offset, final long value, final int scale)
```
RichardWarburton: Is this the most efficient way to do this? Can't we just check if the number of zeros is > 0?

RichardWarburton:
I've added a few inline comments.

artio-codecs/src/test/java/uk/co/real_logic/artio/fields/DecimalFloatToStringTest.java
ariksher-nex: I left a comment so you could see it.  :-)
**must** it be consistent?
look at line below I use Float.valueOf.

artio-codecs/src/main/java/uk/co/real_logic/artio/util/MutableAsciiBuffer.java
RichardWarburton: In that case it's probably better to just remove the scale=19 entry.

artio-codecs/src/test/java/uk/co/real_logic/artio/util/DecimalFloatEncodingTest.java
RichardWarburton: I would just remove this entry. A value of 0 is always equal to zero and there's point in printing out "0.00" when "0" would suffice.

artio-codecs/src/test/java/uk/co/real_logic/artio/fields/DecimalFloatToStringTest.java
RichardWarburton: I think it should be consistent, you don't want to see a value in your debugger that is different from the one printed out on the buffer. The implementation of toString() does not have to be fast though as people should never use that on the critical path of their application.

artio-codecs/src/main/java/uk/co/real_logic/artio/util/MutableAsciiBuffer.java
ariksher-nex: I tried to minimize the code change.
so i treated the leading zero similar to the minus sign.
if going to refer the zeros count it would lead to larger code change.
why do you think it is not efficient? the access to the array is O(1).

RichardWarburton: Well, given that you had the number of zeros in a local field, It just looked to me you could do a quick check on that. Perhaps I'm miss-reading the diff ;) I don't think your approach is bad, just wondering whether it wasn't going the other route?

ariksher-nex: well I managed to figure it out your way. take a look at new code.

RichardWarburton: Great, thanks.

artio-codecs/src/test/java/uk/co/real_logic/artio/fields/DecimalFloatToStringTest.java
ariksher-nex: i tried to make it consistent, than encountered other test that fails...
test class **DecimalFloatToAndFromDoubleTest** method **shouldConvertFromDoubleAccurately**

> uk.co.real_logic.artio.fields.DecimalFloatToAndFromDoubleTest > shouldConvertFromDoubleAccurately FAILED
>     org.junit.ComparisonFailure: expected:<[].01> but was:<[0].01>
>         at org.junit.Assert.assertEquals(Assert.java:115)
>         at org.junit.Assert.assertEquals(Assert.java:144)
>         at uk.co.real_logic.artio.fields.DecimalFloatToAndFromDoubleTest.shouldConvertFromDoubleAccurately(DecimalFloatToAndFromDoubleTest.java:45)

as I understand the code, the consistency between double/float of java lang and DecimalFloat is **more important** than consistency of DecimalFloat to MutableAsciiBuffer.
if that is correct the test of DecimalFloatToStringTest.java might be obsolete.

ariksher-nex: ok made it consistent. they both output with leading zeros.
had to modify test class **DecimalFloatToAndFromDoubleTest** accordingly.

artio-codecs/src/main/java/uk/co/real_logic/artio/util/MutableAsciiBuffer.java
ariksher-nex: removed entirely this block.

artio-codecs/src/test/java/uk/co/real_logic/artio/util/DecimalFloatEncodingTest.java
ariksher-nex: ok removed it.

artio-codecs/src/main/java/uk/co/real_logic/artio/util/MutableAsciiBuffer.java
ariksher-nex: I removed the entire block, with simpler implementation.

artio-codecs/src/test/java/uk/co/real_logic/artio/util/DecimalFloatEncodingTest.java
ariksher-nex: ok, removed it.

artio-codecs/src/test/java/uk/co/real_logic/artio/fields/DecimalFloatToStringTest.java
RichardWarburton: Thanks.

RichardWarburton:
@ariksher-nex - change looks good to me now, but the build is failing on CI - looks like checkstyle.

ariksher-nex:
@RichardWarburton fixed the checkstyle.

PR #232, opened 2019-01-25
When encoding message, use BeginString from header

zamhassam:
When encoding message, use BeginString from header (which now can be overriden). This allows using same encoders for multiple FIX versions, with 'session' layer enforcing correct header values when sending the message.
For backward compatibility when header is constructed / reset, the field is set to default value from dictionary.


RichardWarburton:
Thanks @zamhassam - looks good.

PR #234, opened 2019-01-28
[Zam/Andrey] Extracted interface for header encoders 

zamhassam:
Allows common session logic to operate on all generated artio code.
The interface is based on session_dictionary.xml (which Artio's FIX engine relies on). We haven't got as far as swapping HeaderEncoder for interface in artio-core - but quick evaluation shows this is doable; if implemented this should allow using multiple FIX dialects from different packages in the same process.

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/builder/Encoder.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/builder/Encoder.java
@@ -55,7 +55,7 @@ public interface Encoder
 
     int messageType();
 
-    Object header();
+    SessionHeaderEncoder header();
 
     void resetMessage();
 }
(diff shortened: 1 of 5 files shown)


RichardWarburton:
Thanks for the PR @zamhassam - the only thing I would say is that it doesn't look like this header handles the situation where some of those fields are missing. That is the case for username and password with FIX 4.2. I think you need to potentially generate implementations for those fields that throw an UnsupportedOperationException and hasUsername() / hasPassword() returns false. I think username and password are the only ones that you're likely to have an issue with.

Does the fix-integration project work ok with this PR?

zamhassam:
@RichardWarburton Yes, we ran fix-integration pack against this branch. Username/password are on Logon message, not the header - aren't they?

We started with idea of building generic FIX header interface, but then code generation gets tricky because of differences between 4.2, 4.4 and FIXT - especially with NoHops(627). What we realised however, is that it would be sufficient to have interface just for the session layer to set things like MsgSeqNum, Sender[Target]CompID etc. Then we can have multiple FIX dialects, each in their own package with their own concrete headers - happily coexisting in the same process, and session can send any of those messages. And if some other header fields (like DeliverToCompID) should be set, it can be done by application on a concrete implementation - because application is naturally coupled to a particular dialect. And even though they are defined on the header, such fields are closer to app than session.

Does it make sense?

RichardWarburton:
You're totally right - that does make sense.

PR #235, opened 2019-01-28
Remove functionality for cleaning up components

tom-smalls:
Remove functionality for cleaning up components that only contain a group.

No tests fail with this removal (and a branched version works in our codebase) and allows for simplifying for the generation code. This is a move towards cleaning up how component interfaces are generated


tom-smalls:
Yes

PR #236, opened 2019-01-29
Duplicate sessionId assignment fix

### LINKED ISSUE #233: [BUG] Engine returns DUPLICATE_SESSION on mass logon
Imagine following scenario with single engine serving multiple libraries: 

S1,S2,S3 -- senderCompId
T -- targetCompId
Framer framer;

1. framer.onInitiateConnection(...,"S1", "T", ) results in framer.sessionContexts.counter == 2 &&  currentlyAuthenticatedSessionIds={1}
2. framer.onInitiateConnection(...,"S2", "T", ) results in framer.sessionContexts.counter == 3 && currentlyAuthenticatedSessionIds={2, 1}
3. framer.onMessage(...sessionId=1...) results in framer.sessionContexts.counter == 2 **<- bug here**
4. framer.onInitiateConnection(...,"S3", "T", ) results in framer.sessionContexts.counter == 3 && sessionContext == SessionContexts.DUPLICATE_SESSION 
 
After step 4 there is no way to initiate session for CompositeKey{localCompId=S3, remoteCompId=T} in current engine as this key sits in framer.sessionContexts.compositeToContext forever (or until reset). 

So the code that causes trouble is: 
```
    void onSentFollowerMessage(
...
    {
        if (messageType == LogonDecoder.MESSAGE_TYPE && recordedSessions.add(sessionId))
        {
            // Ensure no future collision if you take over as leader of the cluster.
            counter = sessionId + 1;
```

counter is updated to arrived sessionId+1 value regardless of if there are other inflight initiated logons. Last arrived "onSentFollowerMessage" spoils counter blocking new sessionId generation.

ashananin:
fixes #233 

- do not decrement SessionContexts.counter so incoming composite keys never clash with currently authenticated sessionIds
- real-logic/fix-integration green (+fixVersion=0.21-SNAPSHOT )

### DIFF
--- a/artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/SessionContexts.java
+++ b/artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/SessionContexts.java
@@ -335,9 +335,6 @@ public class SessionContexts
     {
         if (messageType == LogonDecoder.MESSAGE_TYPE && recordedSessions.add(sessionId))
         {
-            // Ensure no future collision if you take over as leader of the cluster.
-            counter = sessionId + 1;
-
             asciiBuffer.wrap(buffer);
             logonDecoder.decode(asciiBuffer, offset, length);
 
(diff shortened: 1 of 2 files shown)


RichardWarburton:
Thanks for your PR @ashananin

PR #248, opened 2019-03-03
MutableAsciiBuffer - support putFloatAscii(final int offset, final long value, final int scale)

ariksher-nex:
in some cases as an output, we do want to leave the given(configured) scale.

while the existing method 
`public int putFloatAscii(final int offset, final DecimalFloat price)`
might ignore the given scale to price due to **DecimalFloat** constructor uses some normalization which might change the given scale.


ariksher-nex:
@RichardWarburton can you please take a look at it?

RichardWarburton:
At the moment we're using `DecimalFloat` to represent decimal numbers. I'm not sure what the value of having this override that takes a pair of primitive values is. Essentially adding more different ways to do the same thing isn't usually a good idea for an API. But maybe there's something I'm missing here?

The DecimalFloat normalisation is to ensure that you have comparable DecimalFloat objects. So I'd be wary of code that throws the pair of scale and value around without having normalised them as this could lead to confusing business logic or incorrect results elsewhere.

I'm confused why its important to output 0.00 rather than 0 - they should both be parse-able as 0 according to the FIX spec, right?

ariksher-nex:
According to FIX spec you are right.

The normalization method itself takes place just for holding 2 values. I think it might affect performance. I can understand if you want to normalize before doing other desired action such as compare.

Our clients require same scale for specific security, so I need to provide it in some way.
Do you have some alternative or way to achieve such?

RichardWarburton:
I see - so your variant of the FIX protocol is defined such that you have to send a decimal float with a certain number of decimal places every time, even if they are always 0?

ariksher-nex:
> I see - so your variant of the FIX protocol is defined such that you have to send a decimal float with a certain number of decimal places every time, even if they are always 0?

yep, i need to output trailing zeros.

RichardWarburton:
Ok, that makes sense as a use-case for the new API method.

Can you also explain what motivated the rewrite of the implementation of the method, with the inline integer blocks, rather than just adding the trailing zeros at the end of the method?

ariksher-nex:
the motivation for changing the implementation is performance.
I see in method `private int putLong(long remainder, final int end)` that Modulo operator being used, which is not considered as an efficient operator.

ariksher-nex:
> Ok, that makes sense as a use-case for the new API method.
> 
> Can you also explain what motivated the rewrite of the implementation of the method, with the inline integer blocks, rather than just adding the trailing zeros at the end of the method?

did as you suggested, support trailing zeros along side the existing implementation.

RichardWarburton:
Great, thanks. Will merge when appveyor is happy.

ariksher-nex:
> Great, thanks. Will merge when appveyor is happy.

back to happy :smiley:

RichardWarburton:
Thanks for the contribution @ariksher-nex

ariksher-nex:
@RichardWarburton , what's the schedule for next release , that includes this PR?

RichardWarburton:
I have one more item to finish off before the next Artio release, so hopefully end of today otherwise tomorrow.

Issue #249, opened 2019-03-06
ResetSequenceNumberCommand resets sequence to 2 instead of 1


alx-ut:
It seems there are a problem in ResetSequenceNumberCommand implementation for not logged session(Step.RESET_RECV).

Actual:

FixEngine.resetSequenceNumber(sessionId) is invoked.

Logon attempt with msgSeqNo=1
Artio sent response:
|8=FIX.4.4|9=124|35=5|49=SRVID|56=interactive2|34=2|50=TENV|52=20190301-13:41:35.869|58=MsgSeqNum too low, expecting 2 but received 1.

Investigation:

ResetSequenceNumberCommand place ResetSequenceNumber message to inboundPublication.

SequenceNumberIndexWriter invokes:
saveRecord(1, resetSequenceNumber.session()); -> updateSequenceNumber(position, 1)

Next acceptor logon:
Framer fills:
lastReceivedSequenceNumber = receivedSequenceNumberIndex.lastKnownSequenceNumber(sessionId)
lastReceivedSequenceNumber = 1. (from previous updateSequenceNumber)

Session#onLogon checks inbound message msgSeqNo to match:
expectedReceivedSeqNum() = lastReceivedMsgSeqNum + 1 (=2)
and reject logon(msgSeqNo=1) with 'MsgSeqNum too low, expecting 2 but received 1' reason.

Possible solution(?):
May be ResetSequenceNumber message should be reset to zero in SequenceNumberIndexWriter:
saveRecord(0, resetSequenceNumber.session()) instead of 1.

PR #250, opened 2019-03-06
Offline sequence reset fix attempt

alx-ut:
Possible fix for
https://github.com/real-logic/artio/issues/249


RichardWarburton:
Thanks for the PR. Can I clarify why you list this as a "possible fix" ?

alx-ut:
Not sure in ResetSequenceNumberCommand#waitSequence variable. May be better to introduce a new Step.
The other think: may be better to keep pair (sessionIndex,sequenceNumber) in SequenceNumberIndexWriter file instead of sequenceNumber only.

RichardWarburton:
Looks like it solves the issue at least, thanks for the PR.

PR #253, opened 2019-03-20
Thread factory for archivingRunner, framerRunner in engine scheduler …

AlexRychkov:
I need to create threads in a custom way with https://github.com/peter-lawrey/Java-Thread-Affinity for bind threads to cores. The only way to do this in affinity is using try-with-resources from target thread.


RichardWarburton:
Thanks for the PR.

I think if we're adding a thread factory to the engine then it would make sense to support it in the LowResourceEngineScheduler as well. The method on `EngineConfiguration` should also have some javadoc to explain its purpose.

PR #255, opened 2019-03-21
All groups to be iterated on components

tom-smalls:
This is a follow up to a pull request made before.

The aim of this change is to move the implementation of repeating group iterators, and provide a method for accessing them, onto the interface of any Component they belong to. Previously, if a repeating group belong to a component then the decoder for that group would move to the generated interface but the iterator would stay in the implementations for said interface. This made code re-use more difficult.

This change involves altering how the Component interfaces are generated. Before all fields for that component (and any fields for nested Components) would be put onto the top level Component's interface; this could no longer be supported due to how Iterators are instantiated (plus it doesn't really make sense exposing fields on aggregates that do not own said fields). Instead, a Component interface will now extend the interface of any nested Components in the same way that an Aggregate with a Component implements the interface for that Component.

This change passes the `fix-integration` tests and is running as a fork in our system

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
@@ -512,11 +512,24 @@ public class DecoderGenerator extends Generator
             {
                 out.append(fileHeader(builderPackage));
 
+                final List<String> interfaces = component.entries()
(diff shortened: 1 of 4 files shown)


RichardWarburton:
Thanks for the PR. This is an API incompatible change, right?

tom-smalls:
Sorry, didn't see your comment.

It is API incompatible when it comes to where an iterator class is defined (imports); I don't *think* the API for getting fields changes

PR #260, opened 2019-04-10
[Tom/Stef] Introduce common interfaces for generated enums.

tom-smalls:
This allows for us to get the FIX representation value for an enum
without having to have a dependency on the generated FIX enums


RichardWarburton:
Thanks, looks good.

PR #263, opened 2019-04-15
Only copy string on read

tom-smalls:
Apologies, this is not the best pull request I have ever created.

This PR introduces the concept of being able to generate decodes that will lazily instantiate string-based values fields if the `FLYWEIGHT_STRINGS` property has been set to true


RichardWarburton:
Thanks for the PR @tom-smalls - I'm going to close this PR as I plan to investigate this a bit on a side branch until I'm happy, but it will definitely get merged in some form.

PR #265, opened 2019-04-20
Adding on @FunctionalInterface to certain interfaces

sh-abhi:
While this is may not be a large change, as I'm going through the code base, I figured I would make some small enhancements as I see them. I've also added on a javadoc to `ReplayProtocolHandler` to make it slightly more straightforward than it already is.


RichardWarburton:
Thanks.

PR #268, opened 2019-04-25
Checksum is wrong for the IndexedPositionWriter in the SequenceNumberIndexWriter

guoguanglei:
SequenceNumberIndexWriter only pass a slice of the underlying buffer to IndexedPositionWriter.

https://github.com/real-logic/artio/blob/585043becbad5372346c4ea59d79c525cce40655/artio-core/src/main/java/uk/co/real_logic/artio/engine/logger/SequenceNumberIndexWriter.java#L109

This will cause the checksum is wrong for the "Positions Table" segment.


RichardWarburton:
Thanks Gary - Makes sense to me.

PR #271, opened 2019-05-11
output windows line endings when generating code on windows

frankwallis:
Because git converts from `\n` to `\r\n` when checking out code to a windows machine, after running the build process all the code-generated files are marked as modified (because they have been rewritten with `\n`).

This change uses `%n` to output the correct line-ending in the code generated file based upon the platform.

It should NOT change runtime behaviour at all, ie. fix messages should still contain `\n` where they did before and that will NOT change depending on platform.

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/AcceptorGenerator.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/AcceptorGenerator.java
@@ -92,8 +92,7 @@ public class AcceptorGenerator
         acceptorOutput.append(String.format(
             "    void on%1$s(final %2$s decoder);%n%n",
             message.name(),
-            decoderClassName(message)
-        ));
+            decoderClassName(message)));
     }
 
     private void generateDefaultAcceptorCallback(final Writer acceptorOutput, final Message message) throws IOException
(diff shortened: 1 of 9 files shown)


RichardWarburton:
Thanks for this PR. I've currently got a branch that I plan to merge in soon that prototypes flyweight based decoders for FIX messages. This alters the generator code somewhat. As a result I would strongly advise you hold off on this change until that gets merged in. It's on the flyweight-codecs branch if you want to take a look.

I was also wondering if there was a better way to get the platform aware newline outputs than convert everything to a printf? Or whether its worth having a test for this, as it looks likes the kind of thing that's easy to introduce a regression around.

frankwallis:
Yes ok I will wait for that branch to be merged and rebase/reapply.

I don't know a better way tbh, you can use `System.lineSeparator()` to get the value for the current platform but this seemed like a lesser impact change.

There are tests in`LineEndingsTests.java` which will fail on the Windows CI build assuming that the code path in question is executed. I spent some time trying to mock the current value of `System.lineSeparator()` but I couldn't get it to work for various reasons. It may be possible with PowerMockito but even then I am not sure.

RichardWarburton:
I've merged in the flyweight codec changes, so you should be able to rebase on top of that now. Sorry for the collisions.

RichardWarburton:
I'm closing this PR due to unfixed merge conflicts. Happy to accept an updated one if you want to progress with this.

PR #277, opened 2019-06-04
Clarify the javadoc for encode and update

clifford:
The parameter accepted by both encode and update, namely epochFraction should have the same time unit as the constructor parameter EpochFractionFormat, to prevent incorrect encoding.


RichardWarburton:
The javadoc change has caused a checkstyle fail btw - https://ci.appveyor.com/project/RichardWarburton/artio/builds/25026514/job/44uqv8w3k8aksk9e

clifford:
Apologies for that will correct it

PR #280, opened 2019-06-26
Add a method in Encoder to accept AsciiSequenceView for String field.

guoguanglei:
A handy method for user who use AsciiSequenceView.


RichardWarburton:
Thanks Gary.

PR #282, opened 2019-06-30
change TcpChannelSupplier from class to interface

alx-ut:
interface is extracted from TcpChannelSupplier.
Existing implementation are moved to DefaultTcpChannelSupplier


RichardWarburton:
Thanks for the PR but being a class enables the JIT compiler to make class hierarchy analysis based optimizations. What's the motivation for this change?

alx-ut:
We need to turn on/off listening socket by schedule and have filtering by IP address.
May be better leave to TcpChannelSupplier as a class and allow to extend listeningChannel behavior via additional policy?

RichardWarburton:
I think it's fine for your to just implement your own TcpChannelSupplier without me doing that. The thing is that if I expose those fields then I'm exposing implementation detail of that class and committing to it for you. If I change the implementation, even for example renaming a field, then I might break your subclass. You're better off to just copy the code a bit more and not coupling an external implementation to this class.

Also - if you just want to filter by IP address you don't need to do it by changing Artio at all - you could just add an `iptables` rule.

alx-ut:
Agree. 
That's the reason of separation between interface(abstract class) and default implementation.

How TcpChannelSupplier can be created without listeningChannel bind in constructor? 
The only solution now - is invoke super() with the specific EngineConfiguration clone without bindAddress.

The copy of the whole class is valid.

iptables exists, but the accepted addresses should should be also controlled by java code.

RichardWarburton:
Maybe an alternative refactor would be to switch the interface to an abstract class? That way you would get the separation that you want but the class hierarchy analysis would also work.

alx-ut:
Agree. 
Change TcpChannelSupplier to abstract class with empty constructor is fine. 
The existing implementation will be moved to DefaultTcpChannelSupplier.
Or it's better to leave existing implementation in TcpChannelSupplier and introduce class AbstractTcpChannelSupplier?

RichardWarburton:
I think better to change the TcpChannelSupplier to an abstract class and add a DefaultTcpChannelSupplier.

PR #284, opened 2019-07-01
change TcpChannelSupplier to abstract class

alx-ut:
Continue of 
https://github.com/real-logic/artio/pull/282


RichardWarburton:
Thanks for the PR. We don't use the @override annotation, inline with the real logic checkstyle guide.

alx-ut:
Annotation is removed

RichardWarburton:
Thanks.

PR #285, opened 2019-07-04
Remove mutable static state in generated enums

tom-smalls:
We came across an issue where we received two FIX messages at the same time on two instances of a FIX Engine running on different threads where one of the messages managed to corrupt the other one when routing based on the MsgType enum. The MsgType enum is the same enum in both FixEngines; when both engines looked up the MsgType enum value, they both used the same mutable CharArrayWrapper field, since this is part of the enum's CharArrayMap.

This PR removes the wrapper from CharArrayMap. Now, a user of CharArrayMap has to pass in an instance of CharArrayWrapper instead. There is now an instance of a CharArrayWrapper for each string-based enum within the aggregate decoder where the enum lives.

As part of this change we stopped special-casing MULTIPLECHARVALUE, which is now treated in the same way as MULTIPLESTRINGVALUE and MULTIPLEVALUESTRING.

This does not introduce any functional change, or garbage, in the critical path. It does represent a breaking change to anyone using the static methods on string-based enums that previously took a char[].


artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
```diff
@@ -456,27 +456,66 @@ private CharSequence validateEnum(final Entry entry, final Writer out)
         }
 
         final String name = entry.name();
-        final String optionalCheck = entry.required() ? "" : String.format("has%s && ", name);
```
RichardWarburton: Why remove the optional check?

tom-smalls:
The optional check was moved rather than removed. The whole validation
method is wrapped in a has now rather than being an online optional check
as wrap has to use the getter for the field rather than the field itself
due otherwise it doesn't work when the decoder is a flyweight

RichardWarburton:
Thanks for clarifying and thanks for the PR.

PR #287, opened 2019-07-22
[Tom] add a check to only write the contents of AsciiSequenceView if there is data to be written.

tom-smalls:
This keeps the functionality inline with what currently happens for
encoding a Charsequence, where an empty charsequence is just ignored

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/EncoderGenerator.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/EncoderGenerator.java
@@ -536,9 +536,10 @@ public class EncoderGenerator extends Generator
             "    }\n\n" +
             "    public %3$s %1$s(final AsciiSequenceView value)\n" +
             "    {\n" +
-            "        if (value.length() > 0)\n" +
+            "        final DirectBuffer buffer = value.buffer();\n" +
+            "        if (buffer != null)\n" +
             "        {\n" +
-            "            %1$s.wrap(value.buffer());\n" +
+            "            %1$s.wrap(buffer);\n" +
             "            %1$sOffset = value.offset();\n" +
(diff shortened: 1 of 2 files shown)


artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/EncoderGenerator.java
```diff
@@ -541,6 +534,16 @@ private String generateStringSetter(
             "        %1$sLength = value.length();\n" +
             "        return this;\n" +
             "    }\n\n" +
+            "    public %3$s %1$s(final AsciiSequenceView value)\n" +
+            "    {\n" +
+            "        if (value.length() > 0)\n" +
```
RichardWarburton: Maybe I've misunderstood something here Tom - but wasn't the problem here the buffer being null, rather than the length being 0? Also - whichever if the fields you pull out it's probably worth putting into a local variable in the generated code.

tom-smalls: Yes it was the buffer being null but as the buffer being null occurs at the same time as length being 0 felt the code would be more readable in this form as it is a bit more explicit about what is checking.

Will move it to a null check if you prefer

RichardWarburton: Yeah, I think the null check would be better. People could subclass AsciiSequenceView and have a null buffer and a non-zero length for example. I think it'll optimise slightly better as well.

tom-smalls: Done

RichardWarburton:
Thanks Tom.

PR #288, opened 2019-07-22
System property for configuring enum values

tom-smalls:
Use the fix.codecs.disable_reject_unknown_enum_value system property
to disable enum value validation. This maintains the default behaviour
of rejecting if an unknown enum value is received but now gives the
option of to opt out of this.

We've found that venues will add new enum values without warning,
which we don't want to reject on. If we don't support the enum value
then we can just ignore it rather than rejecting the whole
fix message

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/builder/RejectUnknownEnumValue.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/builder/RejectUnknownEnumValue.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2015-2017 Real Logic Ltd.
+ *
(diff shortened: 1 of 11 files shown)


tom-smalls:
Wasn't sure if it was best to do this as runtime system property or at compile time

RichardWarburton:
Looks good. I think runtime is fine because with a static final field then its possible to optimize it out. Thanks for the PR.

RichardWarburton:
Looks good. I think runtime is fine because with a static final field then its possible to optimize it out. Thanks for the PR.

tom-smalls:
I have one more commit for this but can't push as GitHub seems to be upset

PR #291, opened 2019-08-05
Option to turn off duplicated fields validation

alx-ut:
We need to turn off duplicated field validation in fix message encoder.
The related example:
http://cdn.batstrading.com/resources/membership/US_Options_FIX_Specification.pdf

ExecutionReport message has fields:
LegRefId,
InstrmtLegExecGrp#LegRefId


tom-smalls:
I don't know how this would affect encoding but sending duplicate fields
definitely breaks the artio decoders when repeating groups are involved as
the decoders use whether they've seen a field before as an indication of
whether are have entered a new entry in a repeating group

RichardWarburton:
Yeah. I mean when you have a field on the message itself and a field on the repeating group and you receive a tag value pair of <field_Id>=<value> it seems pretty ambiguous as to whether that's a value that should be assigned to the message itself or the repeapting group. I don't necessarily mind changing the decoders to be more robust, but it's not really clear what the right behavior even is.

I've had conversations with a couple of clients and venues where the venue has essentially agreed that the duplicate field is an error in their FIX dictionary and should just be removed. This might be the situation in your case.

tom-smalls:
@RichardWarburton When you come across a field that is repeated in a schema in a message, you would probably have to take into account the field seen before and the field that appears after to work out whether you're still in the repeating group or not.

We've come across this before too but because we didn't care about the repeated field we just removed it from the FIX spec of the venue and artio will just skip it.

RichardWarburton:
Hmm, special casing these kinds of fields. Have you actually hit examples of situations where this kind of duplication isn't actually an error?

tom-smalls:
It depends what you mean by error :)

I've come across venues who have intentionally added the same field in two places as they didn't want to introduce a new field, and then realised the problems it introduced...

Do you know if QuickFix handles this case?

alx-ut:
The generated code supports duplicated fields in the message body and repeating group. 
I wrote test. Encoder and Decoder with duplicated fields(message body + repeating group) works fine.
It seems the validation for the required duplicated fields works wrong(it's not necessary).

The example:
<message name="ExecutionReport" msgtype="8" msgcat="app">
...
<component name="Instrument" required="Y"/>
<field name="LegRefID" required="N"/>
<component name="InstrmtLegExecGrp" required="N"/>
...
</message>

<component name="InstrmtLegExecGrp">
            <group name="NoLegs" required="N">
                <component name="InstrumentLeg" required="Y"/>
....
                <field name="LegRefID" required="N"/>
...
            </group>
</component>

RichardWarburton:
Thanks, but there's just `...` in your example message @alx-ut .

alx-ut:
> Do you know if QuickFix handles this case?

It seems the quickfixj allow to have duplicated tag in message body and repeating group.
Today I write more precise tests

tom-smalls:
Having a test where you have InstrmtLegExecGrp followed immediately by LegRefID from Instrument to see if it is all decoded correctly would be good, along with some standard decoding tests would be good.

Trying to think what else we came across when creating https://github.com/real-logic/artio/pull/145 and https://github.com/real-logic/artio/pull/165

alx-ut:
The test with ExecutionReport#LegRefID  followed immediately after InstrmtLegExecGrp is fail(quickfixj test also is fail).
The test with ExecutionReport#LegRefID  before InstrmtLegExecGrp is successful (quickfixj test also pass).

Actually the fix message have ether LegRefId in message body or InstrmtLegExecGrp. Not both simultaneously(it depends on ExecutionReport#ExecType).
So the requirement; field should only appear in a message once is satisfied(for concrete fix message, not schema)

Generally duplicated fields is not valid.  
The case: duplicated field after repeating group can not be parsed unanimously. 
But we need to support client fix spec(encoder + decoder).  
The duplication validation turn off flag is disabled by default, 
It seems it's allowed to have duplicated field in message body/repeating group when field is before repeating group. 
It's not require to change decoder.

artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/CodecGenerationTool.java
```diff
@@ -31,6 +31,8 @@
 
 public final class CodecGenerationTool
 {
+    public static final String VALIDATION_ALLOW_DUPLICATES = "fix.validation.allow.duplicates";
```
RichardWarburton: The format for our other validation properties is `fix.codecs.some_descriptive_name`, so this should be `fix.codecs.allow_duplicate_fields`

artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/DictionaryParser.java
```diff
@@ -443,7 +443,10 @@ private void identifyDuplicateFieldDefinitionsForMessages(final List<Message> me
 
         if (errorMessage.length() > 0)
         {
-            throw new IllegalStateException(errorMessage.toString());
+            if (!Boolean.getBoolean(CodecGenerationTool.VALIDATION_ALLOW_DUPLICATES))
+            {
+                throw new IllegalStateException(errorMessage.toString());
```
RichardWarburton: You should also update the error message in order to tell users that this option can be used in order to switch the duplicate field validation off.

RichardWarburton:
I see. That makes a lot more sense now. I'm happy to merge this PR on that basis, but I've added a couple of comments inline on the PR.

tom-smalls:
> The test with ExecutionReport#LegRefID followed immediately after InstrmtLegExecGrp is fail(quickfixj test also is fail).
> The test with ExecutionReport#LegRefID before InstrmtLegExecGrp is successful (quickfixj test also pass).
> 
> Actually the fix message have ether LegRefId in message body or InstrmtLegExecGrp. Not both simultaneously(it depends on ExecutionReport#ExecType).
> So the requirement; field should only appear in a message once is satisfied(for concrete fix message, not schema)
> 
> Generally duplicated fields is not valid.
> The case: duplicated field after repeating group can not be parsed unanimously.
> But we need to support client fix spec(encoder + decoder).
> The duplication validation turn off flag is disabled by default,
> It seems it's allowed to have duplicated field in message body/repeating group when field is before repeating group.
> It's not require to change decoder.

Would it be worth committing some of these tests? Worth documenting the "dangers" of this feature somewhere perhaps?

RichardWarburton:
+1 to committing the tests and an explanatory doc.

artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/CodecGenerationTool.java
alx-ut: done

artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/DictionaryParser.java
alx-ut: Done

PR #300, opened 2019-10-11
allow compile time alternative for RejectUnknownEnumValue

motocodeltd:
after recent discussion on gitter, we modified the code generator to support an alternative to RejectUnknownEnumValue.CODEC_DISABLE_REJECT_UNKNOWN_ENUM_VALUE_PROP ("fix.codecs.disable_reject_unknown_enum_value").  Providing a boolean value to -Dreject.unknown.enum.value when running the code generator will use that value (resulting in an if (true) / if (false) in the generated code rather than if (CODEC_DISABLE_REJECT_UNKNOWN_ENUM_VALUE_PROP). This allows for compile time configuration without deprecating the old configuration style. Happy to apply this to other configuration if you like this approach.

n.b. we had issues running the fix integration tests locally (with and without our changes), but we've integrated the branch with our code at TransFICC and our tests are passing.

Thanks,

Tom / Stef

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/CodecGenerationTool.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/CodecGenerationTool.java
@@ -67,6 +67,12 @@ public final class CodecGenerationTool
             DECODER_PACKAGE,
             PARENT_PACKAGE);
 
+        final boolean newMode = false;
+        final String newModeValue = "true";
+        final String codecRejectUnknownEnumValueEnabled;
+
+        codecRejectUnknownEnumValueEnabled = fetchRejectionMode(newMode, newModeValue);
+
         final EncoderGenerator encoderGenerator = new EncoderGenerator(
             dictionary,
             ENCODER_PACKAGE,
@@ -74,7 +80,8 @@ public final class CodecGenerationTool
             encoderOutput,
             Validation.class,
             RejectUnknownField.class,
-            RejectUnknownEnumValue.class);
+            RejectUnknownEnumValue.class,
+            codecRejectUnknownEnumValueEnabled);
 
         final DecoderGenerator decoderGenerator = new DecoderGenerator(
             dictionary,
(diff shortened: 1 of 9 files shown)


RichardWarburton:
I've got a set of changes on a branch to support multiple FIX versions that I would like to ensure are merged before any changes to the CodecGenerationTool happen. I expect to have time to merge them in next week.

motocodeltd:
sure let us know and we can rebase our changes. btw the build is failing with the same integration test failures we're getting locally - but we're getting those without our changes too. The build seems to be failing recently
https://ci.appveyor.com/project/RichardWarburton/artio/history

RichardWarburton:
Thanks for pointing that out. I'll investigate the integration tests in a bit, I would expect it to fail on CI for most of the lifetime of 0.38-SNAPSHOT because we were depending upon an agrona snapshot which CI's configuration can't deal with. Additionally those CI tests end up getting run on under-resourced machines most of the time sadly so they aren't as trustworthy as I would like.

RichardWarburton:
I've merged in the multiple fix versions branch btw and fixed the issue with the fix-integration project. So I'm happy to take another look at this when its merge-able.

I like the idea of having the configuration that can be runtime configurable or forcibly switched off / on. Can you write a short explanation that can be used as documentation for this idea and what motivates it? I'm only thinking like a paragraph or so.

motocodeltd:
Hi Richard,

Thanks for notifying us. I've rebased the changes onto the latest artio master. Unfortunately there's still some CI issues - 

```java.nio.file.FileSystemException: logs\sequence_numbers_received: The process cannot access the file because it is being used by another process.
```
in https://ci.appveyor.com/project/RichardWarburton/artio/builds/28202157/job/ixc8dxu6hbbbd7qu/tests


Worth bearing in mind that this change is additive - i.e. if you currently configure artio via a system property you can continue to do so.

The reason why we like this change is, configuring rejections via a runtime property can represent some risk. If this is done programmatically (via System.setProperty), we need to make sure that the class that sets the property is loaded before artio. If this is done at startup by passing a -D to the jvm, this is harder to test and ensure, especially as startup args are often different between different environments such as dev and production (e.g. memory settings). A compile time option solves this issue by ensuring the decoders are generated to ignore unknown enums.

motocodeltd:
Hi @RichardWarburton - any update on this? We believe the failure is not due to our change.Thanks

RichardWarburton:
@motocodeltd thanks - I've merged in the change. I think it's a reasonable idea. I've been on holiday for the last week, thus the delay.

PR #303, opened 2019-11-04
Conditionally generate fix dictionary

motocodeltd:
At Transficc, our Fix dictionaries mostly contain business messages and we generate our session messages separately. The introduction of FixDictionaryGenerator.java breaks our generated code as it adds imports and intitialisations for encoders and decoders that don't exist. This PR conditionally generates those based on if the messages exist, similar to the approach already taken for UserRequest:

```
        final boolean hasUserRequest = dictionary
            .messages()
            .stream()
            .anyMatch(message -> message.name().equals("UserRequest"));
```

Our approach expands this to work generically for all message types.


motocodeltd:
Integration tests fail because they don't test against latest snapshot version. See [fix integration tests pull request.](https://github.com/real-logic/fix-integration/pull/9)

RichardWarburton:
PR makes sense, thanks.

PR #304, opened 2019-11-05
Allow Message Types of up to 5 chars

motocodeltd:
* The problem:
We came across a major venue that uses custom messages with a message type of 5 characters. While artio can read the characters off the wire without a problem, the message type is converted into an integer by packing the first two characters in the lower 16 bits of the 32 bit int. As it happens, our custom messages both started with the same two letters, resulting in a duplicated value in a generated switch statement. It's worth pointing out that there's nothing in the fix spec that we can find, that says that message types have a maximum size of 2.

* The solution:
We changed the way converting a message type into an int works, by using 6 bits to represent any alphanumeric character in the message types. We use 6 bits rather than the ASCII 7 so that we can pack 5 characters (6 * 5 = 30 bits) rather than 4 (4 * 7 = 28 bits) and we do this by not using the ASCII numberical value for each character but a custom charset that excludes symbols etc.

* Side effects:
  * Hardcoded ints in the codebase, such as '0' no longer work, as '0' no longer maps to ASCII 48. Our generator needs to be used. This also means that switching using the constants as case values doesn't work, as the compiler can't resolve them at compile time, so an if statement has to be used.
  * If those numbers are serialised to disk, then the mapping will no longer work between versions because we have changed it. New mapping is, however, stable and two way.
  * We've introduced an exhaustive test that checks all possible alphanumerical permutations of strings sized 1-5 to make sure that our algorithm works. We found this useful as it took a few attempts to get right. The test does take 1.5 mins on our work station though.

This is a significant change, but as long as the ints are used we struggle to see how larger message types could be supported. We could support up to 4 by using the ASCII encoding but we have a concrete example that uses 5.

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/SessionConstants.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/SessionConstants.java
@@ -15,6 +15,8 @@
  */
 package uk.co.real_logic.artio.dictionary;
 
+import static uk.co.real_logic.artio.dictionary.generation.GenerationUtil.packMessageType;
+
 public final class SessionConstants
 {
     public static final byte START_OF_HEADER = 0x01;
@@ -40,17 +42,6 @@ public final class SessionConstants
     public static final int PASSWORD = 554;
     public static final int NEW_PASSWORD = 925;
 
-
-    public static final int LOGON_MESSAGE_TYPE = 65;
-    public static final int HEARTBEAT_MESSAGE_TYPE = 48;
-    public static final int TEST_REQUEST_MESSAGE_TYPE = 49;
-    public static final int RESEND_REQUEST_MESSAGE_TYPE = 50;
-    public static final int REJECT_MESSAGE_TYPE = 51;
-    public static final int SEQUENCE_RESET_MESSAGE_TYPE = 52;
-    public static final int LOGOUT_MESSAGE_TYPE = 53;
-
-    public static final int USER_REQUEST_MESSAGE_TYPE = 17730;
-
     public static final String LOGON_MESSAGE_TYPE_STR = "A";
     public static final String HEARTBEAT_MESSAGE_TYPE_STR = "0";
     public static final String TEST_REQUEST_MESSAGE_TYPE_STR = "1";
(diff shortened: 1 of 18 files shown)


motocodeltd:
```
uk.co.real_logic.artio.integration_tests.SessionMessageValidationTest 
```
is failing for two reasons:
1. it uses the old int representation. I'll submit a PR to fix this
2. even if that is fixed, some of the messages are invalid for other reasons. For example line 32
```
8=FIX.4.4^A35=0^A34=2^A49=TW^A52=<TIME>^A56=ISLD^A999=HI^A
```
has no sequence number and is rejected for that reason. I don't really understand how this is passing in artio master - it's failing for locally for the same reason.

motocodeltd:
See PR https://github.com/real-logic/fix-integration/pull/10

Contrary to what I said earlier, ``uk.co.real_logic.artio.integration_tests.SessionMessageValidationTest`` is passing on my local workstation, but is failing on another due to missing sequence number on some of the test messages.

RichardWarburton:
I'm just looking over this PR now. I appreciate this is exceedingly clever work gents but I'm wondering if it wouldn't be simpler to switch the message type over to be a long and just continue to pack bytes in? That way we could fit 5 character message types in. It would still be a very simple conversion - push 5 bytes into a long like we're currently doing with two bytes and an int.

The only issue is that it would involve a file format change for the FixMessage header, but as far as I can tell this PR doesn't deal with that issue anyway. If someone had an archive of Artio FixMessage SBE entries and this PR was merged - it wouldn't deal with reading the old messageType int values from those historical messages anyway would it? What would happen if you upgraded Artio and tried to replay messages from before the upgrade?

I'm not sure if that last aspect is a killer - but I'd have to solicit opinions from clients as to whether it was something they're ok with.

motocodeltd:
Hi Richard,

Thanks for taking the time to look into this. Yes, using a long would work. We considered it, but came to the conclusion that this is a larger change (and also has some side effects, such as not being able to use a switch statement (provided the value is available at compile time anyway)). The generated code does use a switch statement on the packed ints, for example. While there is some complexity in using 5 bits per character, the exhaustive test makes sure that it's correct.

Yes, if this change (or a change to use a long) is accepted, any persisted messages wouldn't be able to be read from disk and replayed without some conversion. Not sure what the alternative is though, if message types longer than two characters are to be supported.

Let us know how you'd like to proceed. A change to use a long would work very well for us too, though we'd take the opportunity to support larger message types by using the 5 bit encoding. The largest example we currently have is 5 chars, but we haven't seen anything in the fix spec that limits it.

tom-smalls:
Hi @RichardWarburton 

Have you had any thoughts around how to proceed with this?

RichardWarburton:
Sorry - busy week. I was hoping that if we switched over to using longs we could just return to having simple packing of bytes in. I'm just wondering if we really need to keep the character packing. I know there's strictly no limit on the number of characters used for message type identification, but a long gives us 8 bytes to work with and we only know of an exchange that uses 5. I appreciate that you would like to keep the packing it in but I'm wondering if it's really worth the complexity / overhead?

tom-smalls:
We would be happy to move from using an int to a long but as @motocodeltd mentioned, this would be a larger change as we would need to replace all the switch statements (potentially to ifs). If you are happy for this then we can try and work on this approach instead?

To try and remove backwards compatibility issues with your customers, we could deprecate the current msgType field in your SBE schema and introduce a new one for long value and use the SBE version to determine what field to read / act on

RichardWarburton:
I'd be happy to go down the message type switch to long route. Adding a new field and deprecating the old one sounds like a good idea as it will help people avoid hitting the compatibility issue.

motocodeltd:
Hi Richard - will be closing this pull request then and work on the long version. Thanks

PR #310, opened 2019-12-02
add the option to defer binding to the network when starting up

dunkymole:
On top of the bind/unbind API recently added, this allows users to spin up the engine in an unbound state. Currently we have to start the engine bound to the interface then unbind.


RichardWarburton:
PR makes sense. The only thing is I think "deferBinding" could be worded more clearly - defer implies it will happen later but if you do nothing with this configuration it never happens. How about "disableStartupBind" or a "bindAtStartup" option with the opposite meaning that defaults to true?

PR #316, opened 2019-12-10
[Tom] Throw exception if sentinel enum value attempted to be encoded.

tom-smalls:
Encoding the enum sentinel values is not something that should happen, and is usually the result of a programming error. Especially for strings, where the sentinel values can lead to a corrupt FIX message, you want to avoid this.

One thing we weren't sure about is whether to treat NULL_VALUE and ARTIO_UNKNOWN differently, silently dropping NULL_VALUE and throwing for ARTIO_UNKNOWN. Alternatively we could silently drop in both cases. What are your thoughts on this?


RichardWarburton:
I think silently dropping NULL_VALUE and throwing for ARTIO_UNKNOWN makes most sense to me. Essentially setting the value for `field` to NULL_VALUE would have the same effect as calling `resetField()`.

artio-codecs/src/test/java/uk/co/real_logic/artio/dictionary/generation/EncoderGeneratorTest.java
```diff
@@ -1029,4 +1075,21 @@ private void setDataFieldLength(final Encoder encoder) throws Exception
     {
         setInt(encoder, "dataFieldLength", 3);
     }
+
+    private static void assertThrows(final ThrowingRunnable runnable, final Class<? extends Exception> thrown)
+    {
```
RichardWarburton: Maybe I'm misreading this, but it looks like this assert would pass if no exception is thrown and only fail if an exception of the throw type is thrown?

Wouldn't it be better to use [Junit's built in support for this](https://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html)

tom-smalls:
Sure

artio-codecs/src/test/java/uk/co/real_logic/artio/dictionary/generation/EncoderGeneratorTest.java
tom-smalls: Yes, I forgot about that functionality; we use JUnit5, which has a different way for asserting on exceptions and could only remember the @Test(Exception.class) way.

I've fixed this up

tom-smalls:
I have made the changes to just drop for NULL_VAL

RichardWarburton:
I was about to merge this PR, but your other PR, that I just merged, deletes setEnum() that this PR uses.

tom-smalls:
Ahhh, ok. Will fix now

tom-smalls:
I've updated the PR to remove the merge conflicts with the other changes.

PR #317, opened 2019-12-12
[Tom/Richard] Fix required enum char encoding.

rhbradford:
We found that for all enum types, except for char-based ones, throw an exception if they do not have a value set when encoding and they are a required field.

This change introduces the same behaviour for char-based enums, where if we write the default value we end up with a corrupted FIX message


RichardWarburton:
Thanks, looks good.

PR #333, opened 2020-02-17
Allow to implement TcpChannelSupplier by class with non artio package.

alx-ut:
TcpChannelSupplier class has package-private visibility for the abstract fields.
It doesn't allow to implement the class in the other project specific package.
The visibility is changed to 'public'


RichardWarburton:
Thanks Alexey!

PR #340, opened 2020-02-21
Changed AsciiBuffer::scan to use exclusive end index.

AndreyNudko:
There was confusion around that, when some callers passed inclusive index while others (including generated decoders) exclusive. This commit settles on using exclusive index, which is more consistent with other commonly used java APIs.


RichardWarburton:
Cheers.

PR #341, opened 2020-02-22
Initial sequence index configuration

alx-ut:
Initial sequence index configuration.


RichardWarburton:
Hi Alexey. Thanks for the PR. There's a subtle race condition here around reads, but I'll fix that locally.

PR #344, opened 2020-03-03
[Tom] have ascii buffer throw a AsciiNumberFormatException

tom-smalls:
Instead of a NumberFormatException as this is a specific failure case.

This is related to https://github.com/real-logic/agrona/pull/171 encoding

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/util/MutableAsciiBuffer.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/util/MutableAsciiBuffer.java
@@ -15,6 +15,7 @@
  */
 package uk.co.real_logic.artio.util;
 
+import org.agrona.AsciiNumberFormatException;
 import org.agrona.DirectBuffer;
 import org.agrona.concurrent.UnsafeBuffer;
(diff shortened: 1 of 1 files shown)


tom-smalls:
I wasn't able to run all the tests locally as an ILink3ContextsTest  is failing

RichardWarburton:
By default that test should be disabled. Seems to pass by default on my machine and on CI. What commandline are you running?

tom-smalls:
I was running `./gradlew build`

RichardWarburton:
Does it pass if you just run `./gradlew` ?

tom-smalls:
It's working on my laptop but it had been failing on my work computer.

I'll double check tomorrow...

RichardWarburton:
I would guess that you had some cached state from a previous run with the iLink3 stuff enabled, which wouldn't be cleared out if you ran `./gradlew build` and not `./gradlew`.

tom-smalls:
Quite possibly.

All I remember is that a ilink3 test was failing due to a
ConcurrentModificationException

RichardWarburton:
Are you running on JDK 11?

PR #348, opened 2020-03-11
make getId public so that it can be used outside artio.

AmitA396:


RichardWarburton:
Makes sense - but we you also rename id() so that it's consistent with the rest of the public API. Sorry - should have made that suggestion on the original PR.

AmitA396:
Made the suggested change. Please review.

PR #349, opened 2020-03-13
eliminate garbage producted in decimal float formatting during appendTo

vdaniloff:


RichardWarburton:
Thanks for the PR @vdaniloff

PR #350, opened 2020-03-19
fix index catchup to work on start

vdaniloff:
1) Only apply the stream id check during normal operations, not startup replay
2) use the recording id on startup, not the aeron session id to lookup the recording
3) change indexedpositionwriter to use the recording id and not the session id
4) initialise the record offsets map on loading file.

found no easy way to find recording by replay session id

### DIFF
--- a/artio-codecs/src/test/java/uk/co/real_logic/artio/fields/DecimalFloatToAndFromDoubleTest.java
+++ b/artio-codecs/src/test/java/uk/co/real_logic/artio/fields/DecimalFloatToAndFromDoubleTest.java
@@ -3,7 +3,6 @@ package uk.co.real_logic.artio.fields;
 import org.junit.Test;
 
 import java.math.BigDecimal;
-import java.util.Locale;
 
 import static java.math.RoundingMode.UNNECESSARY;
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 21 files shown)


vdaniloff:
looks like a test fails, will continue

artio-codecs/src/test/java/uk/co/real_logic/artio/fields/DecimalFloatToAndFromDoubleTest.java
```diff
@@ -31,40 +32,39 @@ public void shouldExtractNaNFromDecimalNumber()
     @Test
     public void shouldConvertFromDoubleAccurately()
     {
-        final StringBuilder builder = new StringBuilder();
```
RichardWarburton: I don't think these decimal float changes look related to the indexer catchup, can you keep them out of the PR please.

vdaniloff: removed. In Russia this test fails :)

artio-core/src/main/java/uk/co/real_logic/artio/engine/PositionSender.java
```diff
@@ -96,4 +96,11 @@ public void close()
     public void readLastPosition(final IndexedPositionConsumer consumer)
     {
     }
+
+    @Override
```
RichardWarburton: We don't use @Override in real logic projects.

artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/FramerContext.java
```diff
@@ -86,9 +86,11 @@ public FramerContext(
             outboundLibraryStreams.dataPublication("outboundPublication"));
 
         sentSequenceNumberIndex = new SequenceNumberIndexReader(
-            configuration.sentSequenceNumberBuffer(), errorHandler, configuration.logFileDir());
+            configuration.sentSequenceNumberBuffer(), errorHandler, recordingCoordinator.outboundRecordingIdLookup(),
```
RichardWarburton: These changes now mean that the RecordingIdLookup class can be used on the Framer and Indexer agents. It's not a class that is designed to be threadsafe. Probably best to pass out separate instances to each thread.

RichardWarburton:
Thanks for the PR @vdaniloff - My only other suggestion is that this PR doesn't have a way to reproduce the original bug.

artio-core/src/main/java/uk/co/real_logic/artio/engine/PositionSender.java
vdaniloff: removed

artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/FramerContext.java
vdaniloff: fixed, thanks!

vdaniloff:
I suspect this might be the reason PersistentSequenceNumberResentRequestSystemTest fails with "shutdownClearly=false", currently struggling with getting it passed on Windows. 

Originally I was more concerned with repeating reload in "extendRecording" case, since there would be multiple replays for same recording (some of that stored by old aeron session id - so potentially huge)

vdaniloff:
I temporary admit my defeat with PersistentSequenceNumberResentRequestTest, and normal graceful shutdown seems to have no functional impact, except for performance in extendRecording case. The rest is fixed.

PR #358, opened 2020-04-17
removed non gc-free collection

glcanvas:


RichardWarburton:
I think it's a good idea to switch this over to be gc-free but I don't see the point of switching to a HashSet when just need a simple arraylist.

PR #366, opened 2020-05-25
fix password cleaner issue

dmtsyganov:
fix password cleaner issue, when body length changes from 2 to 3 characters, UpdateBodyLengthField throws an exception (from buffer's putNaturalPaddedIntAscii)


RichardWarburton:
Thanks for the PR.

PR #375, opened 2020-07-02
allow more tolerant time-parsing

vdaniloff:
we have some parthers, who remove trailing zeroes in time fields. while that's non-compliant with FIX spec, it still can be parsed anyway. Added a configuration parameter to turn this on/off


RichardWarburton:
Looks good, thanks.

PR #376, opened 2020-07-02
reduce logging during replay operation

vdaniloff:
Replay operation/session may log same state switch a lot of times, sometimes making logs fully unreadable.
This PR only logs changes of the state.


RichardWarburton:
The current Replay logging isn't really designed to be switched on permanently in production - it's for debugging. I don't think production systems should have it enabled.

vdaniloff:
so, should we close then? actually ok with it :)

RichardWarburton:
Yeah - I think that's best as long as you're ok with it.

PR #377, opened 2020-07-06
Fix epoch fraction: convert time to unit of time used in creation

markina:


RichardWarburton:
Build fails checkstyle.

PR #380, opened 2020-07-08
accept exceeding gap range in resend

vdaniloff:
Some partners exceed the range we expect with new sequence. Generally it can be seen artio does not reject this, and stops awaiting a resend, however in this case seq num stays where it was at the resend request.


RichardWarburton:
Looks like the new test failed on CI: https://ci.appveyor.com/project/RichardWarburton/artio/builds/33979273/job/3brh1bh8s1t7ho59/tests

vdaniloff:
thanks, investigating (passes locally)

PR #381, opened 2020-07-09
extend unbind operation with optional disconnect and logout

vdaniloff:
Actually I think EndOfDay is already an out of date name for a command, so probably this has to be renamed as well.

### DIFF
--- a/artio-core/src/main/java/uk/co/real_logic/artio/engine/FixEngine.java
+++ b/artio-core/src/main/java/uk/co/real_logic/artio/engine/FixEngine.java
@@ -102,8 +102,23 @@ public final class FixEngine extends GatewayProcess
      */
     public Reply<?> unbind()
     {
-        return framerContext.bind(false);
+        return unbind(false);
     }
+
+    /**
+     * Unbinds the acceptor socket, and disconnects currently connected TCP connections if requested.
+     *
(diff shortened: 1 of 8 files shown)


RichardWarburton:
Thanks for the PR @vdaniloff .

The only thing that I'm not so sure of here when reviewing this code is what happens if you call unbind and then bind immediately after one another. This didn't used to be a problem because the entire `unbind()` operation within the Framer was only one step on its duty cycle - so unbind and bind wouldn't race. Now unbind has multiple steps because its waiting for the logouts to happen.

I'm not 100% sure what the right behaviour here is either.
1) Would we just want the bind operation to throw an exception?
2) Should it wait until every existing connection is logged out and then rebind the backing port.
3) Should it stop the unbind?

I think (3) seems really hard to reason about - if some sessions got logged out and others didn't. So I think either (1) or (2) is best. But I think we should be clear about the semantics and not end up with a race condition.

vdaniloff:
Good question! Actually I am also not sure what would be expected, 1) is definitely easier to implement, so I will add this.

vdaniloff:
done, if ever need to switch to 2) then polling the pending reply can be added as a step in unitofwork.

RichardWarburton:
Thanks - I'm merging this in - I'll cleanup the success() statements myself and I've also realised we've got some similar undefined behaviour with two unbind operations that I'll fixup.

PR #382, opened 2020-07-09
remove redact after sent reject to be compliant with FIX spec

vdaniloff:
Per FIX specification "Increment inbound MsgSeqNum" step should still be done after sending a reject, increment should not be done only in Logon case.


RichardWarburton:
Thanks Vladimir.

PR #385, opened 2020-07-10
send logout after rejected logon message

vdaniloff:
Per FIX specification correct handling of invalid logon is (Optional) Reject message referencing Logon, sending Logout message and disconnecting

### DIFF
--- a/artio-core/src/main/java/uk/co/real_logic/artio/session/SessionParser.java
+++ b/artio-core/src/main/java/uk/co/real_logic/artio/session/SessionParser.java
@@ -201,6 +201,7 @@ public class SessionParser
             {
                 session.logoutAndDisconnect();
             }
+            return action;
         }
         else if (messageType == LOGOUT_MESSAGE_TYPE)
         {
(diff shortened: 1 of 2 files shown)


artio-core/src/main/java/uk/co/real_logic/artio/session/SessionParser.java
```diff
@@ -195,7 +196,11 @@ private Action rejectExceptionalMessage(final long messageType, final int refTag
     {
         if (messageType == LOGON_MESSAGE_TYPE)
         {
-            return onExceptionalMessage(logon.header(), refTagId, position);
+            final Action action = onExceptionalMessage(logon.header(), refTagId, position);
+            if (action != ABORT)
+            {
+                session.logoutAndDisconnect();
+            }
```
RichardWarburton, line 203: This block should return the action, rather than fall-through. Otherwise in the case of back-pressure the first time you try to send the reject it will fall-through to onExceptionalMessage() call at the end of the method.

PR #387, opened 2020-07-13
mark message as invalid and increment processed seqnum after reject

vdaniloff:
Some of this was lost during redact removal


RichardWarburton:
Thanks @vdaniloff - looks good.

PR #389, opened 2020-07-16
Fix #388

### LINKED ISSUE #388: Engine based messages always have millisecond resolution
Hi, messages sent via the engine itself always have millisecond resolution, because millis-based EpochClock is used. Also, messages sent by a Session#send(Encoder) also have the sending time set to millis-based clock (to support micros one has to call prepare, then set it externally and then send a buffer instead of encoder).

I see several possible fixes for this:
1) switch to EpochNanoClock for that (only used by i3link for now). Possible problem is their subtle discrepancy with millis-based clock, probably this is negligible
2) introduce an EpochFractionClock similar to aeron-cluster ClusterClock. Get rid of epochnanoclock and epochclock in configuration, with only passing one (and then converting)

Happy to fix this myself, but let's discuss it first.

vdaniloff:


artio-core/src/main/java/uk/co/real_logic/artio/session/Session.java
```diff
@@ -49,17 +55,44 @@
 import static uk.co.real_logic.artio.LogTag.FIX_MESSAGE;
 import static uk.co.real_logic.artio.builder.Validation.CODEC_VALIDATION_DISABLED;
 import static uk.co.real_logic.artio.builder.Validation.CODEC_VALIDATION_ENABLED;
-import static uk.co.real_logic.artio.dictionary.SessionConstants.*;
+import static uk.co.real_logic.artio.dictionary.SessionConstants.HEARTBEAT_MESSAGE_TYPE_CHARS;
```
RichardWarburton, line 58: Does this PR need to have so much import statement churn in it?

artio-codecs/src/main/java/uk/co/real_logic/artio/util/CompositeEpochClock.java
```diff
@@ -0,0 +1,61 @@
...
+        this.clock = clock;
+        this.nanoClock = nanoClock;
+    }
+
+    public long time()
+    {
+        return clock.time();
+    }
+
+    public long epochFractionTime(final EpochFractionFormat format)
```
RichardWarburton, line 44: So the way that this implementation works we end up doing the format switch on every single encode. Wouldn't it be better from an efficiency point of view to do the switch once when we start up the program. So have nanos, millis and micros implementations and then a factory method that returns the correct implementation. Sorry - I don't think I described this terribly well on the previous issue comments.

artio-core/src/main/java/uk/co/real_logic/artio/session/Session.java
vdaniloff: unfortunately that's my configuration of optimize imports, I'll revert this

artio-codecs/src/main/java/uk/co/real_logic/artio/util/CompositeEpochClock.java
vdaniloff: Ok, let's rework to constructor parameter rather than method parameter. Need to rethink injection of this into session then.

vdaniloff:
will create separate branch

PR #401, opened 2020-08-27
[Tom/Richard] Carry on decoding a message for field that does not belong at the top/message level

rhbradford:
When decoding a message that has an unexpected tag at the top level of the message that is defined
in a repeating group, we should not stop decoding the message (if we are not failing for unknow field)
but skip that tag and continue with the next one.

There are examples of venues that break the FIX spec and will send tags defined for a repeating group
(in their spec!) outside of the group. In these cases we just want to skip the tag and continue as usual.

If CODEC_REJECT_UNKNOWN_FIELD_ENABLED is enabled then the message will stop decoding


RichardWarburton:
Can we have a unit test for this please? Currently there's pretty good coverage over the codecs.

tom-smalls:
Yes of course.

We wrote tests in our codebase and forgot to add any here. Will sort this
out today.

There is still a case where if you were partway through decoding a
repeating group and came across a field defined elsewhere in the message,
the codec stops trying to decode the repeating group (even if it wasn't
completed) because this is how artio detects the end of a repeating group.
This is obviously a bit pathological but wanted to mention it

tom-smalls:
@RichardWarburton Added a test

RichardWarburton:
Cheers Dude.

PR #403, opened 2020-09-14
Fix edge-case in UtcTimestampEncoder date-roll logic

AndreyNudko:
Because current time was compared with start of next day using `>` rather than `>=`, date-roll executed exactly on midnight (highly unlikely event) was still encoding previous day.

As a side effect, initialize() is now completely optional - date-roll logic would kick in and encode the right date in all cases; which I think is a slight improvement in usability. Please note that for almost all timestamps that was the case already, the only exception was start of epoch which happens to be handled now.
I guess initialise() still can be used as a kind of optimisation to encode the date part during startup.


RichardWarburton:
Looks good - thanks.

PR #409, opened 2020-11-02
Use the same variable for the constructor as used in the classname so they match

anotherchrissmith:
We've received FIX data dictionaries where all the message names are in lower case. Whilst the generation of the java classes handled this without a problem, the constructor names were being derived from a different source and were all in lower case (resulting in compiler errors)


RichardWarburton:
Thanks Chris

PR #410, opened 2020-11-04
Capitalise enum classes

writeoncereadmany:
We were given a data dictionary where enum names were all lowercase, which causes compilation errors in the encoder & decoder classes.  This change ensures the first character of enum names is in uppercase, so calling static methods on the enum from the encoder/decoder classes doesn't clash with the enum field name.


RichardWarburton:
Thanks for the PR.

PR #415, opened 2021-01-01
FIX ResendRequest may cause infinite loop when BeginSeqNo>lastMsgSeqNo

alx-ut:
It seems like an infinite loop might occur in the ReplayQuery#query.
Client send ResendRequest(beginSeqNo,endSeqNo)  with beginSeqNo > when maximum session seqNo.
The cycle in the SessionQuery#query:
while (iteratorPosition != stopIteratingPosition) {..}
will not stop because
offset(iteratorPosition, capacity) 
always point to the some existing record in the index file.


RichardWarburton:
Hi Alexey - thanks. I think this should be safe. It is a circular buffer, but we effectively mod the iteratorPosition value with the capacity in the `ReplayQuery.offset()` method to calculate the offset within the buffer.

PR #416, opened 2021-01-04
artio will freeze during startup when stream_id of cached recording_i…

alx-ut:
Artio is frozen when stream_id for the recordingId in the 'recording_coordinator' was differ from an actual one.
Exception in the engineCounters:
java.lang.IllegalStateException: Unable to reuse recordingId: 1. Stream id is mismatch: actual: 7314, expected: 7313
	at uk.co.real_logic.artio.engine.RecordingCoordinator.acquireRecording(RecordingCoordinator.java:289)
	at uk.co.real_logic.artio.engine.RecordingCoordinator.track(RecordingCoordinator.java:220)
	at uk.co.real_logic.artio.protocol.Streams.dataPublication(Streams.java:71)
	at uk.co.real_logic.artio.engine.EngineContext.inboundPublication(EngineContext.java:343)
	at uk.co.real_logic.artio.engine.framer.FramerContext.<init>(FramerContext.java:84)
	at uk.co.real_logic.artio.engine.FixEngine.initFramer(FixEngine.java:306)
	
May be better to get rid of the recording_coordinator ? It seems to be able to find an entry suitable for extension using listRecordingsForUri?


RichardWarburton:
Thanks - that makes sense to error.

PR #417, opened 2021-01-05
ErrorHandler customization is added

alx-ut:


RichardWarburton:
Hi, I don't think it's a good idea to support two different ways of configuring custom error handling.

Is this really something that you really need? If so then I think it's best to keep one configuration option and adapt when / where it is called?

alx-ut:
The immediate reporting of all errors is desired. It seems the statistical collection may be not enough in some cases.

May be customError handler should be made in addition with the default one:
```
 errorHandler =
                (throwable) ->
                {
                    if (customErrorHandler != null)
                       customErrorHandler.onError(throwable);
                    if (!distinctErrorLog.record(throwable))
                    {
                        System.err.println("Error Log is full, consider increasing " + ERROR_BUFFER_LENGTH_PROP_NAME);
                        throwable.printStackTrace();
                    }
                };
```
or with the recorded flag to the customHandler:
```
boolean recorded = distinctErrorLog.record(throwable);
customErrorHandler.onError(throwable,recorded);

```

I don't find a good way to invoke ErrorConsumer immediate from the distinctErrorLog.record without agrona update.
The invocation may be looks like:
recordId=distinctErrorLog.record(throwable)
ErrorLogReader.read(..recordId)
The direct access is not available because clock.time() is invoke within distinctErrorLog#record.

Or may be better to introduce additional class ErrorHandleStrategy(incompatible changes).
CommonConfiguration.errorHandleStrategy(ErrorHandleStrategy strategy);
interface ErrorHandleStrategy {
 void initialize(MonitoringFile monitoringFile, CommonConfiguration cfg);
 void  nError(Throwable) ;
}

PR #423, opened 2021-03-16
wait for slow image when handing session over

vdaniloff:
if engine session falls back to slow (e. g. fragmented message), then remaining part can be lost, since outbound tracker is switched to other image.


vdaniloff:
same tests failed in https://ci.appveyor.com/project/RichardWarburton/artio/builds/38225703

RichardWarburton:
Thanks @vdaniloff

PR #427, opened 2021-05-19
fix data races in session handover

vdaniloff:
- restored awaitIndexer lost in refactoring
- capturing sequences only after receiver endpoint is paused. 
- catchup scheduling postponed till handover units of work are done (note that RequestOfflineSession also used the value before it is written by checkLoggerUpToDate unit of work is executed


RichardWarburton:
Thanks for the spot and the fix @vdaniloff .

PR #429, opened 2021-06-04
respond to resend request with a too high seq number

vdaniloff:
a resend request with sequence number too high is not answered immeditely as invalid. since it's an admin message it will not be resent from counterparty (a gap fill will be sent instead) - therefore it is never answered

additionally, out-of-sequence, but valid responses are now distinguished from invalid ones (probably could fix without this, but probably still a good idea).

### DIFF
--- a/artio-core/src/main/java/uk/co/real_logic/artio/library/OnMessageInfo.java
+++ b/artio-core/src/main/java/uk/co/real_logic/artio/library/OnMessageInfo.java
@@ -20,7 +20,8 @@ import uk.co.real_logic.artio.messages.MessageStatus;
 public class OnMessageInfo
 {
     private MessageStatus status;
-    private boolean isValid;
+    private boolean valid;
+    private boolean outOfSequence;
 
     public OnMessageInfo status(final MessageStatus status)
     {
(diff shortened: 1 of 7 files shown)


artio-core/src/main/java/uk/co/real_logic/artio/library/OnMessageInfo.java
```diff
@@ -30,7 +31,13 @@ public OnMessageInfo status(final MessageStatus status)
 
     public OnMessageInfo isValid(final boolean valid)
     {
-        isValid = valid;
```
RichardWarburton, line 33: Please don't rename variables unrelated to the change in question, especially when it reduces the consistency of variable naming.

artio-core/src/main/java/uk/co/real_logic/artio/library/OnMessageInfo.java
```diff
@@ -30,7 +31,13 @@ public OnMessageInfo status(final MessageStatus status)
...
+        this.valid = valid;
+        return this;
+    }
+
+    public OnMessageInfo isOutOfSequence(final boolean isOutOfSequence)
+    {
+        this.outOfSequence = isOutOfSequence;
```
RichardWarburton, line 40: Please don't rename variables unrelated to the change in question, especially when it reduces the consistency of variable naming.

RichardWarburton:
Hi @vdaniloff ,

Thanks for the PR I think the FIX behaviour change here is a good idea but I don't think that separating out the "out of sequence" from "is message" is a good idea. Surely it's just a simpler model for people to program against with just one thing to check and understand, rather than two?

regards,

  Richard

vdaniloff:
thanks for review, I've done it in other way in https://github.com/real-logic/artio/pull/432

PR #432, opened 2021-06-08
Respond to resend request with high seq num

vdaniloff:
a resend request with sequence number too high is not answered immeditely as invalid. since it's an admin message it will not be resent from counterparty (a gap fill will be sent instead) - therefore it is never answered


RichardWarburton:
Thanks @vdaniloff - looks good!

PR #436, opened 2021-06-25
reset sequence numbers hangs in sole library mode

vdaniloff:
in sole library mode there's still a gateway session, no session inside. the command then hangs in npe


RichardWarburton:
Hi @vdaniloff ,

In general these bug fixes are great and I'll merge this one without a blocker but I think in order to avoid them being accidentally regressed in future reproducing tests would be really appreciated. Because this subtle point in the code isn't something that I think someone would spot when making a future change to that command.

Issue #441, opened 2021-08-16
Avoid Java warnings around invoking unqualified methods called "yield" in generated encoders/decoders


writeoncereadmany:
More recent versions of Java introduce warnings when invoking unqualified methods called "yield":

```
warning: 'yield' may become a restricted identifier in a future release
yield(item);
^
(to invoke a method called yield, qualify the yield with a receiver or type name)
```

Our build process fails on warnings, and builds encoders written using the CodecGenerator.

Those encoders include a copyTo method, which has patterns like:

```
if (hasExecID())
{
    encoder.execIDAsCopy(execID.byteArray(), 0, execIDLength);
}
```

where the methods are based on the fix tags.

FIX dictionaries often contain a fix tag called "yield". Where that's on a message, this yields (pardon the pun): 

```
if (hasYield())
{
    encoder.yield(yield());
}
```

This code triggers the warning above, which means our build fails when updating to more recent versions of the JDK.

The pattern also exists in xxxDecoder.toEncoder(xxxEncoder). I haven't seen it appear anywhere else but I haven't done a comprehensive search.

One simple fix is - if it were instead:

```
if (hasYield())
{
    encoder.yield(this.yield());
}
```

it would not raise the warning.

RichardWarburton:
Hi Tom,

Thanks for the feedback. It's a good point, I like the simple solution you suggested. Are you planning on sending a PR in for this?

I do plan to add Java 17 to the CI matrix a its an LTS release, which should help catch these kind of issues.

writeoncereadmany:
Yep would be happy to submit a PR, but wanted to check you were happy with proposed approach before diving into the codebase. 

This showed up on Java 16, which is already in the CI matrix - I suspect that this is just a missing test? Which would be unsurprising because it's a real corner case, that the domain just happens to be exposed to.

My PR will include a replicating example, so that'll be covered here on out :)

RichardWarburton:
Yeah - you're right. There should be a test for this case.

writeoncereadmany:
that was surprisingly easy to test!

writeoncereadmany:
Fix merged

PR #442, opened 2021-08-21
#441 qualify generated methods

writeoncereadmany:
see https://github.com/real-logic/artio/issues/441

this-qualify all generated method calls to raw names, as they may be called "yield" (not unusual in a fix dictionary!) which generates compiler warnings in java versions 14 and up

also add a field called "yield" to ExampleDictionary, which led to it failing (prior to the generator changes) due to compiler warnings in the ToEncoderDecoderGeneratorTest


RichardWarburton:
Thanks Tom, looks good.

Issue #444, opened 2021-09-07
Validation fails when an optional component of a required group is absent from message since 0.101 release


writeoncereadmany:
On upgrading from 0.100 to 0.101, we're seeing a lot of message validations on generated decoders fail. We're seeing this in validation blocks like:
```
        {
            int count = 0;
            final LegsGroupIterator iterator = legsGroupIterator.iterator();
            for (final LegsGroupDecoder group : iterator)
            {
                count++;                if (!group.validate())
                {
                    invalidTagId = group.invalidTagId();
                    rejectReason = group.rejectReason();
                    return false;
                }
            }
            if (count != iterator.numberFieldValue())
            {
                invalidTagId = 555;
                rejectReason = 16;
                return false;
            }
        }
```
(note: there is no `if (hasNoLegsGroupCounter)` check before this block)
The implementation of `iterator.numberFieldValue()` has changed between 0.100 to 0.101 from:
```
        public int numberFieldValue()
        {
            return parent.hasNoLegsGroupCounter() ? parent.noLegsGroupCounter() : 0;
        }
 ```
to
```
        public int numberFieldValue()
        {
            return parent.hasNoLegsGroupCounter() ? parent.noLegsGroupCounter() : MISSING_INT;
        }
```
(where `MISSING_INT` is `Integer.MIN_VALUE`)
meaning that we're failing the count check.

This is not the case for all optional repeating groups: most repeating groups are correctly putting the `if (hasNoXXXGroupCounter)` check in place before the validation block. The difference appears to be that the failing repeating groups are specified in the data dictionary as optional components of required groups
```
<messages>
    <message name="TradeCaptureReport" msgtype="AE" msgcat="app">
        fields...
        <component name="TrdInstmtLegGrp" required="N"/>
    </message>
</messages
...
<components>
    <component name="TrdInstmtLegGrp">
        <group name="NoLegs" required="Y">
            fields...
        </group>
    </component>
</components>
```

whereas we're seeing the guards being properly applied to optional components of optional repeating groups:

```
<messages>
    <message name="TradeCaptureReport" msgtype="AE" msgcat="app">
        fields...
        <component name="Stipulations" required="N"/>
    </message>
</messages
...
<components>
    <component name="Stipulations">
        <group name="NoStipulations" required="N">
            fields...
        </group>
    </component>
</components>
```

This means we're now (incorrectly?) rejecting messages as invalid. 

[...]

PR #445, opened 2021-09-08
#444 replicating test of validation of message

writeoncereadmany:
as discussed in #444

ok, it wasn't _that_ hard to put a replicating test together.

verified: this test fails before the change applied in #444, and passes following the change.

also verified (but not included: I can if there's value in it?) that this way of constructing a message is consistent with what's generated by a fix dictionary like my motivating example.


RichardWarburton:
Thanks Tom.

PR #451, opened 2021-11-20
[Java] Fix position in the destination array at which putIntAscii/putLongAscii encode number.

### LINKED ISSUE #450: uk.co.real_logic.artio.util.CharFormatter#putIntAscii and putLongAscii ignore `index` when value is zero or a MIN_VALUE
Invoking `uk.co.real_logic.artio.util.CharFormatter#putIntAscii` with `0` or `Integer.MIN_VALUE` produces wrong result as the value encoded at the wrong offset (i.e. zero) in the `buffer` instead of the specified offset (i.e. `index`).

The `uk.co.real_logic.artio.util.CharFormatter#putLongAscii` has the same issue with `0` and `Long.MIN_VALUE`.

vyazelenko:
Fixes #450.

As a bonus it also fixes encoding of the `Long.MIN_VALUE`. ;)


RichardWarburton:
Thanks @vyazelenko

Issue #452, opened 2021-11-23
It is possible to redefine numeric constants such as DecimalFloat.ZERO


michaelszymczak:
I have noticed that some known predefined DecimalFloat values are prone to be corrupted if assigned directly.

![image](https://user-images.githubusercontent.com/7302886/143026685-cb384b18-f381-4cd0-8383-19cfd0ecf982.png)

```java
final DecimalFloat currentPrice;
currentPrice = DecimalFloat.ZERO;
assertThat(currentPrice.toString()).isEqualTo("0");

// When
currentPrice.set(new DecimalFloat(5));

assertThat(DecimalFloat.ZERO.toString()).isEqualTo("0");
```

As shown above, this is especially dangerous in the case when someone uses this constant to set the initial value of some field/variable in their codebase. If this initial value is subsequently updated, it affects all usages and zero becomes a non-zero value.
Also, providing a static mutable fields that looks like constants is probably misleading.

Can we discuss how we can make it less dangerous to use?

vyazelenko:
The same is true of other constants such as `DecimalFloat#NAN` which is among other things is used to denote a missing value when parsing from ASCII (i.e. [uk.co.real_logic.artio.builder.CommonDecoderImpl#getFloat](https://github.com/real-logic/artio/blob/b297ea77f643c2758bc1c464b9fbacc5397a45f2/artio-codecs/src/main/java/uk/co/real_logic/artio/builder/CommonDecoderImpl.java#L83-L104)). In the parsing case once the parsing failed and the same flyweight is used again the `DecimalFloat#MISSING_FLOAT` will be mutated as a result of a successful parsing.

RichardWarburton:
I think making them immutable is a good idea, you're right it does reduce the scope for errors here. I'm not sure what the best approach is mind.

One choice would be to introduce a parent class to DecimalFloat, eg: ImmutableDecimalFloat, that has the accessors but no mutating methods on. Constants could be changed to use this immutable version. It would be backwards incompatible at the API level but I think only in cases where a bug could happen, eg: calling a setter on DecimalFloat.NaN.

I'm open to other suggestions if there's a better option.

vyazelenko:
Another way to solve it would be to add an `immutable` flag which could only be set via a private constructor. This way all constants would be declared with `immutable=true` while normal constructor will create mutable objects. Then you would have to add a check to every mutator method to ensure that object is mutable and throw otherwise, e.g.:
```java
[...]

PR #454, opened 2022-02-06
Issue452/prevent modifying decimal float constants

michaelszymczak:
This change fixes the issue #452 

The purpose of this change is to avoid subtle bugs when the seemingly constant value is redefined.

The change tried to conform to the following constraints:

- maintain as much backward compatibility as possible with a minimal or no change required by the library user

- do not adversely affect the performance

### Design

The approach here is to extract a read only superclass. It has several benefits.

- both types can be used interchangeably, so the client's code should just work without them knowing that anything changed in most cases
- the tandem ReadOnly and the mutable DecimalFloat as still effectively final (cannot be extended outside of the library package)
- there should not be any performance overhead (illegal operations are now checked and rejected at compile time)
- it supports immutability for constants that are still visible as if they were declared on the existing DecimalFloat class
- improved discoverability of potential bugs (as above illegal operations are now checked and rejected at compile time)

### Migration process

Upgrade the Artio dependency, compile the downstream project. If no compilation errors, no changes required.

I have tested this version with a large scale project that makes heavy use of Artio codecs and this version was backward compatible apart from one place that took me 30 seconds to change (passing a ZERO as a mutable input argument which is arguably better off being passed as a ReadOnlyDecimalFloat instead)

When it comes to the backward compatibility, the new version should work just fine.
the only required change on the client side (that should be quite rare anyway) is in places where constant such as ZERO or NAN were assigned or passed  to a function that expected DecimalFloat.

The client can either:
- change expected DecimalFloat type to ReadOnlyDecimalFloat
- ask for a mutableCopy()

There should be no other changes required (comparison, equality etc. maintains backward compatibility and works fine with both types interchangeably)

### Value for library users

If there are other changes required, e.g. invocation of a mutating method such as setValue no longer compiles, it means that the upgrade detected a potential error in the client's codebase that tried to modify a constant value. This is the very reason the change was introduced.

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/fields/ReadOnlyDecimalFloat.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/fields/ReadOnlyDecimalFloat.java
@@ -133,7 +133,6 @@ public class ReadOnlyDecimalFloat implements Comparable<ReadOnlyDecimalFloat>
         CodecUtil.appendFloat(builder, this);
     }
 
-    @Override
     public String toString()
     {
         final StringBuilder builder = new StringBuilder();
@@ -141,7 +140,6 @@ public class ReadOnlyDecimalFloat implements Comparable<ReadOnlyDecimalFloat>
         return builder.toString();
     }
 
-    @Override
     public final int compareTo(final ReadOnlyDecimalFloat other)
     {
         final long value = this.value;
@@ -193,9 +191,7 @@ public class ReadOnlyDecimalFloat implements Comparable<ReadOnlyDecimalFloat>
      * Transitive : if x.equals(y) return true and y.equals(z) return true, x.equals(z) return true
      * Consistent : invocations of x.equals(y) return the same value
      * For any x != null, x.equals(null) returns false
-     * <p>
-     * Any class other than DecimalFloat inheriting ReadOnlyDecimalFloat should overwrite this equal method.
-     * <p>
+     *
      * Support for DecimalFloat was done to provide a backward compatibility when some constants such as ZERO or NAN
      * were made read only to avoid subtle bugs. By making DecimalFloats equal the corresponding ReadOnlyDecimalFloat
      * any code that used the predefined constants when they were mutable should still yield the same results,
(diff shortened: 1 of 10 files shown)


michaelszymczak:
I think Java 11 windows job is just flaky as it failed several times before

artio-codecs/src/main/java/uk/co/real_logic/artio/fields/ReadOnlyDecimalFloat.java
```diff
@@ -0,0 +1,295 @@
...
+    {
+        return this.scale;
+    }
+
+    public void appendTo(final StringBuilder builder)
+    {
+        CodecUtil.appendFloat(builder, this);
+    }
+
+    @Override
```
RichardWarburton, line 136: We don't use `@Override` in Artio code.

artio-codecs/src/main/java/uk/co/real_logic/artio/fields/ReadOnlyDecimalFloat.java
```diff
@@ -0,0 +1,295 @@
...
+     * "Obey the general contract when overriding equals",
+     * and it should do it for every x and y where x and y can be an instance of
+     * ReadOnlyDecimalFloat or DecimalFloat, in any order, this method is:
+     * Reflexive : x.equal(x) return true
+     * Symmetric : if x.equals(y) return true, y.equals(x) return true
+     * Transitive : if x.equals(y) return true and y.equals(z) return true, x.equals(z) return true
+     * Consistent : invocations of x.equals(y) return the same value
+     * For any x != null, x.equals(null) returns false
+     * <p>
+     * Any class other than DecimalFloat inheriting ReadOnlyDecimalFloat should overwrite this equal method.
```
RichardWarburton, line 197: This method is final - is that possible?

michaelszymczak: Good spot, I contradicted myself. I think at some point I realised that we actually don't want anyone to extend this class and forgot to delete the comment.

RichardWarburton:
Cool, thanks

PR #457, opened 2022-04-08
Do not match subsequences when matching headers

wojciech-adaptive:
Header predicate for foobar will match foo, but I believe that should be exact.

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/CodecUtil.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/CodecUtil.java
@@ -148,7 +148,7 @@ public final class CodecUtil
         final int expectedOffset,
         final int length)
     {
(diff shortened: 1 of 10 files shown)


RichardWarburton:
Hi Wojciech,

I think the bug you've found is a good spot but shouldn't it be fixed in the following method in `CodecUtil`?

```public static boolean equals(
        final char[] value,
        final char[] expected,
        final int offset,
        final int expectedOffset,
        final int length)```

RichardWarburton:
Because you've fixed it for the predicate here but the same problem exists elsewhere.

wojciech-adaptive:
I considered that, but it seemed like this method is more generic and can compare arbitrary subsequences, so there's no way to get expected length unless we add a new parameter.

Looking again at this method, shouldn't it be using offsets in the length check?

RichardWarburton:
Yeah, I think it should.

RichardWarburton:
Thanks.

PR #458, opened 2022-04-12
adding sessionId and firmId long support to forNextSessionVerID

vitor-tadashi:
Changing sessionId and firmId in forNextSessionVerID method to support long


RichardWarburton:
Thanks Vitor.

PR #461, opened 2022-05-21
Add @Generated annotation to generated types

mchenry-adaptive:
This PR adds a `@Generated` annotation to generated source types. This is useful for various tools (e.g. code coverage analysis) which can use the annotation to omit the generated code from reports.

The annotation `uk.co.real_logic.artio.dictionary.Generated` is being added, rather than using `javax.annotation.Generated` because:
1. The Java annotation was not available until Java 9 (and some still use java 8)
2. The Java annotation has only SOURCE retention, which is not available to code analyzers that work on class files.


RichardWarburton:
Thanks @mchenry-adaptive PR looks fine, I'll merge it once the CI has gone through.

PR #479, opened 2023-03-31
Added notApplied method to the binary connection interface.

renanleonellocastro:
The purpose of this patch is to add the possibility for the user to send a notApplied message to the client using the already created proxy notApplied method. It was added an notApplied method to the binary connection interface. This method was implemented in the internalBinaryConnection class and calls the proxy notApplied method. This way the user can manually send a notApplied message to the client when it needs it.


wojciech-adaptive (commented):
Looks OK. What do you think about renaming the method to `trySendNotApplied` to be explicit that we're sending a message and that it might not succeed?

PR #482, opened 2023-04-26
Customisable cancel on disconnect behaviour

BrentDouglasB1:
This changes providing a customizable cancel on disconnect option to an engine that will be applied in the case a login message does not provide one. The use case if for supporting an api that always uses cancel on disconnect and does not want to provide an option in the FIX api.

### DIFF
--- a/artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/FixGatewaySessions.java
+++ b/artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/FixGatewaySessions.java
@@ -89,6 +89,8 @@ public class FixGatewaySessions extends GatewaySessions
     private final int forcedHeartbeatIntervalInS;
     private final boolean disableHeartbeatRepliesToTestRequests;
     private final boolean isReproductionEnabled;
+    private final CancelOnDisconnectOption cancelOnDisconnectOption;
+    private final int cancelOnDisconnectTimeoutWindowInMs;
 
     // Initialised after logon processed.
     private SessionContext sessionContext;
@@ -137,6 +139,8 @@ public class FixGatewaySessions extends GatewaySessions
         this.resendRequestController = configuration.resendRequestController();
         this.forcedHeartbeatIntervalInS = configuration.forcedHeartbeatIntervalInS();
         this.disableHeartbeatRepliesToTestRequests = configuration.disableHeartbeatRepliesToTestRequests();
+        this.cancelOnDisconnectOption = configuration.cancelOnDisconnectOption();
+        this.cancelOnDisconnectTimeoutWindowInMs = configuration.cancelOnDisconnectTimeoutWindowInMs();
 
         sendingTimeEncoder = new UtcTimestampEncoder(epochFractionPrecision);
     }
@@ -203,6 +207,8 @@ public class FixGatewaySessions extends GatewaySessions
         session.closedResendInterval(gatewaySession.closedResendInterval());
         session.resendRequestChunkSize(gatewaySession.resendRequestChunkSize());
         session.sendRedundantResendRequests(gatewaySession.sendRedundantResendRequests());
+        session.cancelOnDisconnectOption(gatewaySession.cancelOnDisconnectOption());
+        session.cancelOnDisconnectTimeoutWindowInNs(gatewaySession.cancelOnDisconnectTimeoutWindowInNs());
 
         final SessionParser sessionParser = new SessionParser(
             session,
@@ -246,7 +252,7 @@ public class FixGatewaySessions extends GatewaySessions
 
         return new FixPendingAcceptorLogon(
             sessionIdStrategy, gatewaySession, logon, connectionId, fixContexts, channel, fixDictionary, framer,
-            remoteAddress, fixReceiverEndPoint);
+            remoteAddress, fixReceiverEndPoint, cancelOnDisconnectOption, cancelOnDisconnectTimeoutWindowInMs);
     }
 
     void onUserRequest(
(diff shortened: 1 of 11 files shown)


BrentDouglasB1:
For any future patches, is there a gradle task or something to apply the format from this repo automatically? I kept the format manually, but presumably you have some tooling for it that I couldn't see?

BrentDouglasB1:
@wojciech-adaptive Could you comment on what the chances are of something like this getting merged?

artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/FixGatewaySessions.java
```diff
@@ -203,6 +205,7 @@ void acquire(
         session.closedResendInterval(gatewaySession.closedResendInterval());
         session.resendRequestChunkSize(gatewaySession.resendRequestChunkSize());
         session.sendRedundantResendRequests(gatewaySession.sendRedundantResendRequests());
+        session.cancelOnDisconnectOption(cancelOnDisconnectOption);
```
wojciech-adaptive, line 208: That doesn't seem right, it will use the default instead of the selected option, plus setting the timeout might be missing too. Would be great to add a test for this.

wojciech-adaptive (changes requested):
The JavaDoc in `CancelOnDisconnectTimeoutHandler` will need updating. Please remember that this change affects initiator sessions too.

wojciech-adaptive:
> For any future patches, is there a gradle task or something to apply the format from this repo automatically? I kept the format manually, but presumably you have some tooling for it that I couldn't see?

I think the closest we have is this: https://github.com/real-logic/aeron/blob/master/config/ide/idea/aeron.xml

artio-system-tests/src/test/java/uk/co/real_logic/artio/system_tests/CancelOnDisconnectSystemTest.java
```diff
@@ -145,9 +168,24 @@ public void shouldNotTriggerCancelOnDisconnectTimeoutWithNoLogonOptions()
...
+    {
+        launch(CancelOnDisconnectOption.CANCEL_ON_DISCONNECT_ONLY, FixDictionaryImpl.class);
+        setupWithoutConnecting(SessionCustomisationStrategy.none());
+        connectSessions(FixDictionaryImpl.class);
+
+        acquireAcceptingSession();
+
+        assertEquals(CancelOnDisconnectOption.CANCEL_ON_DISCONNECT_ONLY, acceptingSession.cancelOnDisconnectOption());
+        assertEquals(CancelOnDisconnectOption.DO_NOT_CANCEL_ON_DISCONNECT_OR_LOGOUT,
+            initiatingSession.cancelOnDisconnectOption());
```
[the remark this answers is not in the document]
BrentDouglasB1: Makes sense. Added this and a few other assertions

BrentDouglasB1:
@wojciech-adaptive I think have addressed all comments, though I am not super happy the javadoc change is particularly clear.

I'm not sure I understand your comment about initiator session though. In the earlier version of this PR I was erroneously using the configuration value for initiator sessions here:
https://github.com/real-logic/artio/blob/master/artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/Framer.java#L1561
however I have fixed that now. Otherwise the only way I see it affecting initiator sessions is that if the acceptor has a FixDictionary that supports COD, they will get the acceptor configured defaults echoed back on the login, rather than the static defaults. Is this what you meant? Did you mean one of these things or is there another way that it affects initiator sessions that I am overlooking?

wojciech-adaptive:
> I'm not sure I understand your comment about initiator session though. In the earlier version of this PR I was erroneously using the configuration value for initiator sessions here: https://github.com/real-logic/artio/blob/master/artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/Framer.java#L1561 and using the default rather than the session value in FixGatewaySessions#aquire, however I have fixed that now. Otherwise the only way I see it affecting initiator sessions is that if the acceptor has a FixDictionary that supports COD, they will get the acceptor configured defaults echoed back on the login, rather than the static defaults. Did you mean one of these things or is there another way that it affects initiator sessions that I am overlooking?

Yeah, COD callbacks aren't a thing for initiators, but they need to specify COD parameters in the Logon message, originally it was done using `SessionCustomisationStrategy`, with your previous change the engine defaults would also affect that. I think either is fine, but we should make sure to cover all cases in JavaDoc and tests. Sticking to `SessionCustomisationStrategy` for now seems good to me.

artio-core/src/main/java/uk/co/real_logic/artio/session/Session.java
```diff
@@ -2633,7 +2634,7 @@ void cancelOnDisconnectTimeoutWindowInNs(final long cancelOnDisconnectTimeoutWin
     {
         this.cancelOnDisconnectTimeoutWindowInNs = Math.min(
             MAX_COD_TIMEOUT_IN_NS, cancelOnDisconnectTimeoutWindowInNs);
-        cancelOnDisconnect.cancelOnDisconnectTimeoutWindowInNs(cancelOnDisconnectTimeoutWindowInNs);
+        cancelOnDisconnect.cancelOnDisconnectTimeoutWindowInNs(this.cancelOnDisconnectTimeoutWindowInNs);
```
wojciech-adaptive, line 2637: Good spot.

wojciech-adaptive (approved):

BrentDouglasB1:
Thanks. I was looking at that before but I don't see how I would edit the wiki. Do I clone it and then somehow open a PR against the wikis repo?

wojciech-adaptive:
My bad, thought anyone can edit the wiki. I'll update it.

PR #484, opened 2023-06-15
Reset buffer references in string-like fields in generated encoders

anotherchrissmith:
The generated encoders contain different ways of setting string-like fields.  Some methods take buffers owned by the application, e.g. passing in an AsciiSequenceView.  Other methods take data that is copied into the internal buffer, e.g. passing in a CharSequence.

If an application uses both styles, it could first pass in an AsciiSequenceView (which would point to some kind of input buffer), then on a later code path, pass in a CharSequence.  This will result in the encoder copying the CharSequence into the input buffer, corrupting that buffer.

This change takes a reference to the buffer created in the UnsafeBuffer (owned by the encoder) whenever the application passes in an application-owned buffer, like an AsciiSequenceView.  On reset, it resets the UnsafeBuffer to point it back to the original buffer.

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
@@ -434,6 +434,17 @@ class DecoderGenerator extends Generator
         }
     }
 
+    protected String resetLength(final String name)
+    {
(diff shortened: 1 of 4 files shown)


anotherchrissmith:
The failing checks look like intermittent test failures.  A different set of tests failed on the first run.

anotherchrissmith:
Superseded by https://github.com/real-logic/artio/pull/485

PR #485, opened 2023-06-27
Prevent writing to application-supplied buffers when encoding strings

wojciech-adaptive:
Hi, @anotherchrissmith. One issue I see in https://github.com/real-logic/artio/pull/484 is that the internal buffer reference could end up pointing to an application-supplied buffer, e.g. if you call any of the setters twice. Here I slightly changed it to save the internal buffer reference only when the encoder allocates a new one. Calling setters of different kinds for the same field could still overwrite a buffer, but after reset it should be fine, and we could add a conditional wrap before copying if needed as it should be unlikely. This should save some GC write barriers too.

Also fixed an unrelated bug which I found when reading the surrounding code.

Can you please test this and let me know if it works for you?

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/CodecUtil.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/CodecUtil.java
@@ -106,18 +106,23 @@ public final class CodecUtil
     }
 
     // NB: only valid for ASCII bytes.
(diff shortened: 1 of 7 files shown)


anotherchrissmith:
Hi @wojciech-adaptive.

Thanks for looking at this.  Good spot on the issue that would have been caused if setters were called twice.  I just replayed the journal from our production issue while using this version of the codecs and it fixed the issue.

I'd be happy for this to be merged into master.

Many thanks,

Chris

PR #488, opened 2023-07-17
Allowing application to request logoutDisconnect in a session and spe…

lucianoviana:
…cify a custom text

### DIFF
--- a/artio-core/src/main/java/uk/co/real_logic/artio/session/Session.java
+++ b/artio-core/src/main/java/uk/co/real_logic/artio/session/Session.java
@@ -201,6 +201,8 @@ public class Session
 
     private DisconnectReason pendingDisconnectReason;
 
+    private byte[] logoutText;
+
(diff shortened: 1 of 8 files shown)


artio-core/src/main/java/uk/co/real_logic/artio/session/Session.java
```diff
@@ -587,6 +587,16 @@ public Reply<ThrottleConfigurationStatus> throttleMessagesAt(
             id, throttleWindowInMs, throttleLimitOfMessages);
     }
 
+    /**
+     * Default version of {@link Session#logoutAndDisconnect(byte[])}
```
wojciech-adaptive, line 591: Would be great to clarify that logout won't have text.

lucianoviana:
i've just pushed again implementing your suggestions - thx

wojciech-adaptive (approved):

PR #497, opened 2023-11-27
Allow disabling validation on empty tags

writeoncereadmany:
We are encountering situations where we're being sent FIX messages with empty tags (eg |448=|). Currently, this fails validation. 

Whilst this is clearly invalid FIX, we can't legislate for what our counterparties send us: we've had multiple counterparties send us malformed FIX in this way. It would be nice to be able to, instead of failing on such messages, instead to treat these tags as missing (still triggering required fields validation if the field is required).

This change replicates the approach taken to other settings allowing enabling/disabling of validations, and includes a minimal test to show that the messages pass validation with absent fields when this specific validation is disabled.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,14 +38,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ github.ref }}
-      - name: Create Release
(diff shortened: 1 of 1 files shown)


tom-smalls:
Is this going to get merged?

PR #509, opened 2024-06-10
Add option to replace binary separator with pipe

Nadia-Adaptive:
A change to FixArchivePrinter to allow replacing the binary separator in the FIX message body with a pipe via program arguments.

### DIFF
--- a/artio-core/src/main/java/uk/co/real_logic/artio/engine/logger/FixArchivePrinter.java
+++ b/artio-core/src/main/java/uk/co/real_logic/artio/engine/logger/FixArchivePrinter.java
@@ -30,6 +30,7 @@ import uk.co.real_logic.artio.messages.FixMessageDecoder;
 import uk.co.real_logic.artio.messages.FixPProtocolType;
 import uk.co.real_logic.artio.messages.MessageStatus;
 
+import java.io.PrintStream;
 import java.util.function.Predicate;
 
 import static java.lang.Long.parseLong;
@@ -50,7 +51,7 @@ public final class FixArchivePrinter
 {
     public static void main(final String[] args)
     {
-        new FixArchivePrinter().scan(args);
+        new FixArchivePrinter(System.out, System.err).scan(args);
     }
 
     private FixPProtocolType fixPProtocolType = FixPProtocolType.ILINK_3;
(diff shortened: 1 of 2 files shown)


vyazelenko:
@Nadia-Adaptive Have you considered adding `--delimiter=X` so that any character can be specified?

artio-core/src/main/java/uk/co/real_logic/artio/engine/logger/FixArchivePrinter.java
```diff
@@ -387,8 +405,8 @@ private static void print(
     {
         final MessageStatus status = message.status();
         final long timestamp = message.timestamp();
-        final String body = message.body();
-        System.out.printf("%1$20s: %2$s (%3$s)%n", timestamp, body, status);
+        final String body = pipeDelimiter ? message.body().replace('\u0001', '|') : message.body();
```
chrjohn, line 408: I don't know how far you would like to take this, but `SOH` could also appear in `data` fields, so you might end up with pipe symbols in fields that should not be replaced.
But I think this approach is good enough for the majority of cases.

wojciech-adaptive:
> @Nadia-Adaptive Have you considered adding `--delimiter=X` so that any character can be specified?

We should implement this.

artio-core/src/main/java/uk/co/real_logic/artio/engine/logger/FixArchivePrinter.java
wojciech-adaptive: Good point, thanks. I think this is good enough for now, `fix.core.debug.separator` doesn't handle data fields either.

artio-system-tests/src/test/java/uk/co/real_logic/artio/system_tests/ArchivePrinterIntegrationTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+
+
+
+
+
+
+
+
+
```
wojciech-adaptive, line 126: formatting

artio-system-tests/src/test/java/uk/co/real_logic/artio/system_tests/ArchivePrinterIntegrationTest.java
```diff
@@ -0,0 +1,126 @@
...
+import static uk.co.real_logic.artio.system_tests.SystemTestUtil.connect;
+import static uk.co.real_logic.artio.system_tests.SystemTestUtil.delete;
+import static uk.co.real_logic.artio.system_tests.SystemTestUtil.launchInitiatingEngine;
+import static uk.co.real_logic.artio.system_tests.SystemTestUtil.newInitiatingLibrary;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+
+import org.junit.Before;
+import org.junit.Test;
```
wojciech-adaptive, line 33: Let's use JUnit 5 for new tests.

wojciech-adaptive: Actually we probably can't due to the base class.

Issue #504, opened 2024-03-18
Samples failed to start due to ConfigurationException


TKYAngus:
When trying to start the sample applications (e.g. ExchangeApplication, SampleClient, SampleServer, etc.) I always got exceptions, e.g.:
```
Exception in thread "main" io.aeron.exceptions.ConfigurationException: ERROR - Archive.Context.controlChannel must be set
```

```
Exception in thread "main" io.aeron.exceptions.ConfigurationException: ERROR - Archive.Context.replicationChannel must be set
```

Can I know are those configs required? If so I think we should fix the sample codes

martin-nginio:
@RichardWarburton and @wojciech-adaptive 

any help getting this sample working would be appreciated.

0x26res:
I got SampleClient/SampleClient to work with the following changes https://github.com/real-logic/artio/pull/516

0x26res:
This is fixed in master with 09b8b13fb247484f16acebd3515bfe73388c46f6 though I haven't tested ExchangeApplication.

martin-nginio:
thanks @0x26res

marc-adaptive:
These should be fixed in 09b8b13fb247484f16acebd3515bfe73388c46f6 and e2891587765af8d2505b3f6e00599ac180bc2a96. Please let us know if you are still having problems.

PR #516, opened 2024-07-23
Fix config for sample client/server

0x26res:
See https://github.com/real-logic/artio/issues/504


marc-adaptive:
Could you please try again with latest from master? 

https://github.com/real-logic/artio/commit/09b8b13fb247484f16acebd3515bfe73388c46f6

0x26res:
I can confirm SampleServer and SampleClient are now working in master. Thanks @marc-adaptive

Issue #494, opened 2023-09-27
Generated encoders / decoders generate warnings when compiled under JDK21


writeoncereadmany:
Generated classes using artio-codecs (including latest snapshot build) generate warnings when compiled under JDK21. We have our build configured to fail on warnings, so this is a blocker to us upgrading to JDK21.

Example:

<path>/HeaderEncoder.java:38: warning: [this-escape] possible 'this' escape before subclass is fully initialized
        beginStringAsCopy(DEFAULT_BEGIN_STRING, 0, DEFAULT_BEGIN_STRING.length);

(not all with the same method or on the same generated class)

This is a new linter warning introduced in JDK21: http://marxsoftware.blogspot.com/2023/03/jdk-21-b15-new-xlint-warnings.html

A few possible approaches:

1) Just suppress the warnings with @SuppressWarnings("this-escape"). I've tested this locally and it works if we apply it to decoder/encoder classes, but it's not very principled.
2) Make generated endoder/decoder classes final. I'm not sure why anyone would want to extend them, but this would be a breaking change. It also highlights some other issues which aren't worth diving into if that's not a path worth pursuing.
3) Make just the methods called from constructors final. This is a touch more involved as it requires identifying what methods can be called from constructors, and is still a breaking change albeit a much more minor one
4) Make any methods called from constructors delegate to a private method, with the constructor calling the private method. That will still allow users to override the method, but not in a way which interferes with construction.

I'm happy to prepare a PR for any of these, my personal preference is 2) as it's the simplest and making it clear that the codecs aren't designed for extension seems like a plus, but that's not really my call to make.

marc-adaptive:
For now we've merged in your PR for approach 1. Please re-open if you'd like to revisit

PR #531, opened 2024-12-04
Suppress "this-escape" warnings from generated encoders and decoders

writeoncereadmany:
See https://github.com/real-logic/artio/issues/494

This is the simplest, least invasive solution to avoid warnings on generated classes.


writeoncereadmany:
We needed various other changes on the branch we use in production: I'll raise separate branches for the two features we want merged, the last commit also has an issue raised

PR #533, opened 2024-12-06
Suppress "this-escape" warnings from generated encoders and decoders

writeoncereadmany:
see https://github.com/real-logic/artio/issues/494

This is the simplest, least intrusive change to allow generated decoders to be built under JDK21 without warnings.


writeoncereadmany:
any objections to this?

lucianoviana:
hi - could you please suggest a dictionary so we can reproduce this warning?

writeoncereadmany:
I've attached the most minimal fix dictionary I could create that reproduces - just one application message with one field.

This yields a warning on HeaderEncoder as a result of the constructor, which calls beginStringAsCopy(), which is an instance method - not permitted (without generating a this-escape warning) under JDK21.

[FIX44-artio.txt](https://github.com/user-attachments/files/18968013/FIX44-artio.txt)

PR #534, opened 2024-12-06
[Java] Permit decoders to be built that do not validate empty tags

writeoncereadmany:
We have seen numerous cases where venues are sending us invalid FIX of the form 35=0|12345=|10-123

ie, where a tag is defined with no value / an empty string. 

This does not conform to the FIX standard, so it's technically correct to reject the message as malformed, however: if a venue's sending us that and is refusing to fix it (or is incapable of doing so), we'd rather _not_ fail validation and instead just treat it as empty.

This change allows a build-time flag on generated decoders to ignore empty tags instead of rejecting them.


artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/CodecConfiguration.java
```diff
@@ -129,6 +131,21 @@ public CodecConfiguration wrapEmptyBuffer(final boolean wrapEmptyBuffer)
         return this;
     }
 
+    /**
+     * Suppresses checks for empty tags, eg |1234=|
+     * instead of rejecting when tag is empty, treat the tag as absent
```
wojciech-adaptive, line 136: Not really, because you're still parsing those fields, the `hasFoo()` methods will return true. I think that's dangerous, because you won't be able to tell a difference between, for example, `|43=N|` and `|43=|`.

writeoncereadmany: Oh, damn, good point. I'll revise to make sure we don't parse the tags either and properly treat them as absent.

writeoncereadmany:
I've now updated the change to not parse the tags, and added tests to assert those fields are absent (rather than present but empty/default). 

I've included both String and non-String field tests too (to show it gracefully handles both parsed and non-parsed fields).

artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/CodecConfiguration.java
writeoncereadmany: I've now updated the change to not parse the tags, and added tests to assert those fields are absent (rather than present but empty/default).

I've included both String and non-String field tests too (to show it gracefully handles both parsed and non-parsed fields).

writeoncereadmany:
any chance we can get this reviewed and merged?

PR #536, opened 2024-12-19
Duty cycle tracker

marc-adaptive:
Add duty cycle trackers to framer, library, indexer


artio-core/src/main/java/uk/co/real_logic/artio/engine/EngineContext.java
```diff
@@ -375,10 +375,16 @@ private void newArchivingAgent()
...
+        final Agent dutyCycleTrackingAgent = new IndexerDutyCycleTracker(
+            configuration.agentNamePrefix(),
+            clock,
+            fixCounters.getIndexerDutyCycleTracker(configuration.indexerCycleThresholdNs()));
+
         final List<Agent> agents = new ArrayList<>();
         agents.add(inboundIndexer);
         agents.add(outboundIndexer);
         agents.add(replayer);
+        agents.add(dutyCycleTrackingAgent);
```
wojciech-adaptive, line 387: While your solution is cleaner, I'd like to avoid changing thread names, can we squeeze that tracking into one of the existing agents?

artio-core/src/main/java/uk/co/real_logic/artio/FixCounters.java
```diff
@@ -149,6 +157,36 @@ public AtomicCounter negativeTimestamps()
         return negativeTimestamps;
     }
 
+    public DutyCycleTracker getFramerDutyCycleTracker(final long threshold)
+    {
+        return new DutyCycleStallTracker(
```
wojciech-adaptive, line 162: Maybe in case threshold is zero we could return `DutyCycleTracker` to allow disabling?

PR #548, opened 2025-06-04
Fix validation bug from #546 causing failure for rest of day

maxgawason:
This PR fixes a bug introduced in #546 where a validation failure would cause every subsequent message to fail validation. #546 added validation for the length of boolean and char fields, asserting that the fix tag was only a single character long. Using `Rule80A` (tag 47) as an example, a buffer would be created for the decoder:
```
this.rule80AAsChars = new char[1];
```
When a message was decoded, the value of the fix tag would be copied into the buffer:
```
case 47:
   this.rule80A = buffer.getChar(valueOffset);
   this.rule80AAsChars = buffer.getChars(this.rule80AAsChars, valueOffset, valueLength);
   break;
```
`getChars` is designed to create a new, larger buffer if the field it attempts to read is too large for the current buffer:
```
public char[] getChars(final char[] oldBuffer, final int offset, final int length)
{
    final char[] resultBuffer = oldBuffer.length < length ? new char[length] : oldBuffer;
    for (int i = 0; i < length; i++)
    {
        resultBuffer[i] = getChar(i + offset);
    }
    return resultBuffer;
}
```
The validation would then check the size of the buffer length:
```
} else if (Validation.CODEC_VALIDATION_ENABLED && this.rule80AAsChars.length > 1) {
```
Because the buffer was never resized/reset for each fix message, anytime large tag value was sent (e.g. `47=AB`) the buffer size would be increased and remain that size. Each subsequent check of `${tag}AsChars.length` would then fail. 

This PR removes the `${tag}AsChars` field and replaces it with a `${tag}Length` field for boolean and char fields that holds the length of the field to ensure subsequent messages with the correct length are valid. It maintains the same validation behavior while avoiding heap allocations for every message.

### DIFF
--- a/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
+++ b/artio-codecs/src/main/java/uk/co/real_logic/artio/dictionary/generation/DecoderGenerator.java
@@ -771,7 +771,7 @@ class DecoderGenerator extends Generator
         {
 
             charsValidation = String.format(
-                "        if (%3$s && %1$sAsChars.length > 1)\n" +
+                "        if (%3$s && %1$sLength > 1)\n" +
                 "        {\n" +
                 "            invalidTagId = %2$s;\n" +
                 "            rejectReason = " + VALUE_IS_INCORRECT + ";\n" +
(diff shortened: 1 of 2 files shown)


lucianoviana:
thx much @maxgawason - we're going to take a look

PR #555, opened 2025-08-22
changing size type to long in MappedFile

lucianoviana:


vyazelenko:
The underlying JDK abstraction is limited to `Integer#MAX_VALUE`. See [java.nio.channels.FileChannel.MapMode,long,long)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/channels/FileChannel.html#map(java.nio.channels.FileChannel.MapMode,long,long)).

PR #556, opened 2025-10-06
Aeron upgrade to 1.49.0 and agrona to 2.3.0

lucianoviana:


gradle/libs.versions.toml
```diff
@@ -1,22 +1,22 @@
...
+checkstyle = "11.1.0"
 commons-codec = "1.15"
 commons-lang3 = "3.8.1"
 hamcrest = "3.0"
 hdrHistogram = "2.2.2"
 httpcore = "4.4.14"
 jmh = "1.37"
 json = "20250517"
-junit = "5.13.2"
+junit = "5.13.4"
```
vyazelenko, line 13: Actually, Aeron upgraded to `6.0.0` in the `1.49.0` release.

vyazelenko: But I see that you still use JUnit 4 tests. So with `6.0.0` there will be a warning about junit-vintage.

lucianoviana: ah yeh - I noticed now I was looking at agrona's deps intead of aeron's. Will check. Thx

PR #558, opened 2025-10-13
Allow empty fix tags

writeoncereadmany:
We have seen numerous cases where venues are sending us invalid FIX of the form 35=0|12345=|10-123

ie, where a tag is defined with no value / an empty string.

This does not conform to the FIX standard, so it's technically correct to reject the message as malformed, however: if a venue's sending us that and is refusing to fix it (or is incapable of doing so), we'd rather not fail validation and instead just treat it as empty.

This change allows a build-time flag on generated decoders to ignore empty tags instead of rejecting them.


marc-adaptive:
Thanks for the PR @writeoncereadmany. I think OtfParsing will be impacted by this change (I'm unsure if we should be giving a callback for an empty field, although I could go either way, and an empty group will result in an error). I'll spend some time this week looking into this.

PR #570, opened 2026-03-28
Release replay query subscriptions when replay operations finish

klittlepage:
From what I can tell, the change introduced in 31e14a6 is necessary, but not sufficient. This change ensures that catchup replay no longer leaves the replay query subscribed after it is idle, preventing stale subscriber positions from accumulating on the replay stream.


lucianoviana:
hi, @klittlepage 

Thx for your help with the pr but we'll not proceed with this as the bug has not been proved to exist yet. You can find more details in the issue comments.

PR #577, opened 2026-06-18
fix: send trailing gap-fill when catchup replay ends with heartbeats

### LINKED ISSUE #576: CatchupReplayer returns MISSING_MESSAGES when replay range ends with trailing heartbeats
### Problem
CatchupReplayer incorrectly returns `SessionReplyStatus.MISSING_MESSAGES` when the replay range ends with heartbeat messages (35=0), even though all messages are present in the archive.

### Root Cause
In `CatchupReplayer.onFragment()`, heartbeat messages (35=0) are handled specially — they are accumulated into a range and skipped with return CONTINUE without updating replayFromSequenceNumber:
```java
if (messageType == HEARTBEAT_MESSAGE_TYPE) {
    if (heartbeatRangeSequenceNumberStart == OUT_OF_RANGE) {
        heartbeatRangeSequenceNumberStart = headerDecoder.msgSeqNum();
    }
    return CONTINUE;  // replayFromSequenceNumber is NOT updated
}
```
The gap-fill mechanism (which sends a SequenceReset-GapFill for the accumulated heartbeat range and implicitly advances the sequence accounting) is only triggered when a non-heartbeat message follows a heartbeat series:
```java
if (heartbeatRangeSequenceNumberStart != OUT_OF_RANGE) {
    if (!sendGapFill()) { return ABORT; }
}
return processNormalMessage(...);  // updates replayFromSequenceNumber
```
When the replay ends with heartbeats (no non-heartbeat message follows), neither replayFromSequenceNumber is updated nor is sendGapFill() invoked for the trailing range. After `ReplayOperation.pollReplay()` returns true, `hasMissingMessages()` checks:
```java
return replayFromSequenceIndex < replayToSequenceIndex || 
       replayFromSequenceNumber < replayToSequenceNumber;
```
Since replayFromSequenceNumber was only advanced up to the last non-heartbeat message, this check incorrectly returns true, resulting in `MISSING_MESSAGES`.

### Reproduction
A new system test `TrailingHeartbeatCatchupReplayTest` demonstrates the issue: https://github.com/artiofix/artio/pull/577

1. Establish a persistent FIX session between two gateways
2. Exchange a business message (TestRequest)
3. Generate trailing heartbeats by sending TestRequests from acceptor to initiator (the initiator auto-responds with Heartbeats that become inbound for the accepting engine)
4. Disconnect without sending Logout (to keep heartbeats as the last inbound messages)
5. Restart engines and request a session replay via `FixLibrary.requestSession()` starting from the last business message
6. **Expected**: `SessionReplyStatus.OK`; **Actual (before fix):** `SessionReplyStatus.MISSING_MESSAGES`

Key detail: if a Logout message follows the heartbeats (which is the normal case for a clean disconnect), the bug is masked because Logout is a non-heartbeat message that triggers `sendGapFill()` for the preceding heartbeat range. The bug only manifests when the session is disconnected without a Logout, or when the last messages before any non-graceful disconnection are heartbeats.

### Impact
Any application using `FixLibrary.requestSession()` with replay (e.g., session handover with catchup) will encounter MISSING_MESSAGES if the session's last recorded inbound messages are heartbeats. Since heartbeats are the most common FIX message during idle periods, this affects any session that was idle before an ungraceful disconnection.

### Fix
PR: https://github.com/artiofix/artio/pull/577

After `ReplayOperation.pollReplay()` completes, check if there is an unfinished heartbeat range (`heartbeatRangeSequenceNumberStart != OUT_OF_RANGE`). If so, send a trailing gap-fill covering the remaining heartbeats before evaluating `hasMissingMessages()`.

### Affected Versions
Verified on 0.178.

PavelSmirn:
CatchupReplayer incorrectly returns SessionReplyStatus.MISSING_MESSAGES when the replay range ends with heartbeat messages (35=0), even though all messages are present in the archive.

See https://github.com/artiofix/artio/issues/576 for more details.

Fixes https://github.com/artiofix/artio/issues/576

### DIFF
--- a/artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/CatchupReplayer.java
+++ b/artio-core/src/main/java/uk/co/real_logic/artio/engine/framer/CatchupReplayer.java
@@ -285,7 +285,7 @@ public class CatchupReplayer implements ControlledFragmentHandler, Continuation
         {
             if (heartbeatRangeSequenceNumberStart != OUT_OF_RANGE)
             {
-                if (!sendGapFill())
+                if (!sendGapFill(headerDecoder.msgSeqNum()))
                 {
                     return ABORT;
                 }
@@ -296,7 +296,7 @@ public class CatchupReplayer implements ControlledFragmentHandler, Continuation
         }
     }
 
-    private boolean sendGapFill()
+    private boolean sendGapFill(final int sequenceNumberEnd)
     {
         if (sequenceResetEncoder == null)
         {
@@ -329,10 +329,8 @@ public class CatchupReplayer implements ControlledFragmentHandler, Continuation
             }
         }
 
-        final int heartbeatRangeSequenceNumberEnd = headerDecoder.msgSeqNum();
-
         sequenceResetEncoder.header().msgSeqNum(heartbeatRangeSequenceNumberStart);
-        sequenceResetEncoder.newSeqNo(heartbeatRangeSequenceNumberEnd);
+        sequenceResetEncoder.newSeqNo(sequenceNumberEnd);
         sequenceResetEncoder.header().sendingTime(
             timestampEncoder.buffer(), timestampEncoder.encodeFrom(nanoClock.nanoTime(), TimeUnit.NANOSECONDS));
 
@@ -343,7 +341,7 @@ public class CatchupReplayer implements ControlledFragmentHandler, Continuation
             encodeBuffer, encodedOffset, encodedLength,
             libraryId, SEQUENCE_RESET_MESSAGE_TYPE,
             messageDecoder.session(), replayFromSequenceIndex, libraryId,
-            CATCHUP_REPLAY, heartbeatRangeSequenceNumberEnd) > 0;
+            CATCHUP_REPLAY, sequenceNumberEnd) > 0;
 
         if (sent)
         {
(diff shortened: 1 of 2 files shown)


PavelSmirn:
Hi @vyazelenko 

A gentle nudge on this one - fixes #576. Mind approving the CI run? Happy to rebase/tweak if anything.
Thanks!

PavelSmirn:
The `shouldGapfillAllOutboundMessagesInFlight` failure on Java 17 Windows seems to be a flaky test. The PR only touches `CatchupReplayer`, not `GapFiller` or the test itself. 

Could you please re-run the failed job?
Thanks!

marc-adaptive:
Thanks for raising the issue and PR. Have you considered the case there are missing messages after a heartbeat. For example an inbound sequence of 1=login, 2=new order single, 3=heartbeat, but requesting 1-10. Will try to look into this issue this week.

PavelSmirn:
Thanks for the feedback. Yes, this is a valid concern.

I’ve updated the fix and added a new test to cover this case.

The algorithm now looks like this:

1. If replay ends with an open heartbeat range, send a gap-fill through the last heartbeat actually replayed (lastHeartbeatSeqNum + 1).
2. Update replay progress to that actual sequence number.
3. Check whether the requested endpoint was reached.
4. Return MISSING_MESSAGES if it was not; otherwise return OK.

PR #578, opened 2026-06-30
check for session id in use when creating new LibraryExtendPosition objects

nbradac:
This should address https://github.com/artiofix/artio/issues/575


vyazelenko:
@nbradac Have you considered using `io.aeron.Aeron#nextSessionId` instead?

nbradac:
@vyazelenko I don't see a 'nextSessionId' anywhere.  Also, for whatever reason, Artio is pulling these session ids out of the space between the reserved low/high values configured on the driver, and it's got its own map of session ids (which it's also using as a 'library id').

vyazelenko:
@nbradac [Aeron#nextSessionId(int streamId)](https://github.com/aeron-io/aeron/blob/1.52.0/aeron-client/src/main/java/io/aeron/Aeron.java#L515) which sends a command to the media driver to fetch next `sessionId` that is guaranteed to be unique for the specified `streamId`.

vyazelenko:
> Also, for whatever reason, Artio is pulling these session ids out of the space between the reserved low/high values configured on the driver, and it's got its own map of session ids (which it's also using as a 'library id').

Using driver configuration properties in Artio makes no sense as those won't be available if driver runs in a standalone process. Even if it runs embedded those configuration values could have been programmatically overridden and therefore accessing them via system properties will result in wrong values being used.

nbradac:
@vyazelenko Ok, just using nextSessionId is much simpler.

